Issue #13, opened 2022-08-06
Sending wallpapers to the daemon twice in quick succession makes all further respones fail


Aiz0:
Using the command below
`swww img wallpaper.jpg && swww img wallpaper.jpg `

results in error:
`Error: "Failed to receive answer: io error: Resource temporarily unavailable (os error 11)"`
and higher than normal memory usage sometimes above 200M

after this, using the command
`swww kill`
then fails with the same error.

I then have to kill the daemon and remove the socket manually.

Horus645:
I also can't reproduce this error. Unlike [#12](12), however, I don't actually know what might be causing this.

In particular, it shouldn't be possible for `swww kill` to fail while the daemon is running.

Can you please answer a few questions?
  1. Are you using the latest version of `swww`?
  2. Did you by any chance send images that need to be resized?
  3. If you wait some time, does the image eventually appear on screen?
  
I am thinking that it might just be another issue of the timeout being too low. You send the image, it takes a while to process it, but the client doesn't wait and fails. You try to kill the daemon but it is *STILL* processing the image, so it fails again.

Aiz0:
Just like #12 it was mostly fixed by updating to the latest version.
The issue still remains with some WebP images though.

This one for example
https://files.catbox.moe/xi5toc.webp
It normally renders correctly and appears on screen without error.

When sent twice in quick succession it doesn't seem to appear at all and `swww kill` fails

Aiz0:
I should probably mention that I use two monitors, one 1440p and one 1080p, so the image does need resizing for one of the monitors.

After some testing with the `--outputs` flag I can only reproduce it when sending to both outputs

It seems that it appears on screen sometimes, but it looks very weird.

![2022-08-06_T21:20:32](https://user-images.githubusercontent.com/66974576/183263295-8539c3d6-fdb4-45b2-9b6b-9c295b51b8f1.png)

Horus645:
That's very interesting. Thank you for the through explanation! This will probably take me a little longer since it seems like a more involved problem than the previous one.

---
Meanwhile, you made me realize that's it's been a while since I did a release, and the current code has a lot more stuff than v0.3.0, so I went ahead and released v0.4.0 with all the newer improvements. I will most likely release 0.4.1 as soon as I manage to fix this.

Horus645:
Hello, I just managed to reproduce it somewhat and made a [PR](https://github.com/Horus645/swww/pull/21) that should have fixed. Could you please test it just to confirm it?

[...]

PR #21, opened 2022-08-20
Fix deadlock when sending images quickly

Horus645:
The problem was that we were using SyncSender to stop the animations.

The we use a syncronous channel for sending the frames from the
animation threads in order to ensure old threads that will terminate
soon and new ones don't overlap when sending their frames. However, this
does block the animation threads when they are sending messages.

Because we were also using a syncronous channel to stop the animations,
it was possible that the animation thread was blocked trying to send in
a frame, while the main thread was blocked trying to send in the stop
message.

This fixes that problem by transforming anim_stoppers into a vector of
async Senders.


Horus645:
Attempts to fix #13.

PR #31, opened 2022-09-23
Add a systemd service file

WhyNotHugo:
This is mostly for distribution packages to include. It should be copied into `/usr/lib/systemd/user/swww.service`.

### DIFF
--- a/contrib/swww.service
+++ b/contrib/swww.service
@@ -1,0 +1,12 @@
+[Unit]
+Description=Solution to Wayland Wallpaper Woes
+Documentation=file:/usr/share/doc/swww/README.md
+Requires=graphical-session.target
+PartOf=graphical-session.target
+After=graphical-session.target
+
+[Service]
+ExecStart=/usr/bin/swww init --no-daemon
+
+[Install]
+WantedBy=graphical-session.target


Horus645:
Thanks once again for your work, but I am not sure I want to merge this. This seems like something distribution packagers should do; not us. The problem is that not everyone uses systemd (though yes, I know most do), and so this will be forcing everyone to download this extra file.

More importantly, though, I myself do not feel comfortable maintaining this service file, since I don't really have any experience in writing them (and, personally, I prefer to start `swww` from an init script, rather than a service).

WhyNotHugo:
This is mostly to avoid the duplicate work for all distribution packagers (e.g.: instead of re-writing the same thing or keeping it downstream, they just copy the upstream file).

> More importantly, though, I myself do not feel comfortable maintaining this service file, since I don't really have any experience in writing them (and, personally, I prefer to start swww from an init script, rather than a service).

Sure, it's really up to you, so feel free to close this if you'd rather not. Packagers can, eventually, just fetch it from this PR if they really want to.

A minor note: just by the fact that the service is installed won't run it automatically; it still has to be explicitly enabled by end users (via `systemctl --user enable swww`).

PR #34, opened 2022-09-23
Render the cursor on an empty desktop

WhyNotHugo:
Wayland clients are expected to render the cursor on their input region. By setting the input region to an empty region, the compositor renders the default cursor.

Without this, and empty desktop won't render a cursor.

### DIFF
--- a/src/daemon/mod.rs
+++ b/src/daemon/mod.rs
@@ -13,7 +13,7 @@ use smithay_client_toolkit::{
             LoopHandle, LoopSignal,
         },
         client::protocol::{wl_output, wl_shm, wl_surface},
-        client::{Attached, Display, EventQueue, Main},
+        client::{protocol::wl_compositor, Attached, Display, EventQueue, Main},
         protocols::wlr::unstable::layer_shell::v1::client::{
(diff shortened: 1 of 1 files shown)


Horus645:
I don't see why that would be a problem. You aren't capable of interacting with `swww` in any way with a cursor, and so if the desktop is otherwise empty, it seems reasonable that the cursor doesn't have to be rendered, no?

WhyNotHugo:
It's mostly confusing. When there's two screens, it's hard to figure out on which of both screens the mouse focus is. It's also tricky to interact with bars or panels on empty desktops.

If you like the idea of actually hiding the cursor on an empty desktop I can move this change behind a separate flag (e.g.: `--show-cursor`).

Horus645:
You know what? I've gotten around this. It's just a few lines of code that can potentially make some people's life easier.

> Wayland clients are expected to render the cursor on their input region. By setting the input region to an empty region, the compositor renders the default cursor.
Without this, and empty desktop won't render a cursor.

May I ask you to just add a comment explaining this in the code itself, for future reference?

WhyNotHugo:
Updated.

PR #36, opened 2022-09-30
many big changes

flick0:
merged `any` `center` `from` transitions into **`grow`**
merged `left` `right` `top` `bottom` transitions into **`wipe`**
changed `outer` to **`shrink`**
renamed `simple` to **`fade`**

added bezier curve interpolation to all transitions (except `fade` ofc)
added ability to run the `wipe` transition with any angle given with `--transition-pos`

## new flags:
+ `--transition-bezier` : bezier curve to use in transitions: `ease-in` | `ease-out` | `ease-in-out` | `a,b,c,d`(bezier curve's control point positions) (defaults to `ease-in-out`)

+ `--transition-pos` : position of transition *(center of circle in case of `shrink` and `grow`, angle of line in case of `wipe`)*<br><br>**`grow`/`shrink`** : `top` | `left` | `right` | `bottom` | `top-left` | `top-right` | `bottom-left` | `bottom-right` | `x,y` *(coords of a point in screen space)*<br><br>**`wipe`** : `top` | `left` | `right` | `bottom` | `angle` *(angle with respect to an imaginary horizontal line running across the center of screen, angle can also be given as `{angle}deg`)* (defaults to `0,0` |  `0deg`)

### grow
grow
### shrink
shrink
### wipe
wipe

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 target/*
 completions/*
 test_images/*
+Cargo.lock
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9,12 +9,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
 
 [[package]]
-name = "adler32"
-version = "1.2.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "aae1277d39aeec15cb388266ecc24b11c80469deae6067e17a1a7aa9e5c1f234"
-
-[[package]]
 name = "assert_cmd"
 version = "2.0.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -72,16 +66,10 @@ dependencies = [
 ]
 
 [[package]]
-name = "bumpalo"
-version = "3.10.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "37ccbd214614c6783386c1af30caf03192f17891059cecc394b4fb119e363de3"
-
-[[package]]
 name = "bytemuck"
-version = "1.11.0"
+version = "1.12.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a5377c8865e74a160d21f29c2d40669f53286db6eab59b88540cbb12ffc8b835"
+checksum = "2f5715e491b5a1598fc2bef5a606847b5dc1d48ea625bd3c02c00de8285591da"
 
 [[package]]
 name = "byteorder"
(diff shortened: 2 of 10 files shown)


Horus645:
Thanks!

This looks pretty cool, and it seems like it was a lot work, which makes me rather reluctant to say this, but this PR is simply way too big. Each one of those changes are really big in and of themselves. Furthermore, I feel like maybe you should have opened an issue for discussion first for some of these, **specially** the CLI changes; those will break everyone's configuration on update. Sure, I suppose not that many people are using `swww` yet, but I don't want to break their configs if I can avoid it.

I would really generally prefer if we had 3 PRs:
  1. One just for the CLI changes
  2. Another one just for the Bezier Curves
  3. Another one just for the doing the `wipe` with an angle.
  
The reason being, all of these have a bunch of different problems we need to address: are the CLI changes really necessary (I feel like they aren't, but I would like to hear your thoughts)? Do we really need to use the `curves` and `keyframe` crates for the Bezier stuff? I would rather keep dependencies to a minimum, if possible. That said, if we *do* go with `keyframe`, could we also use it for the other animations? *Should* we do so? And then, what would be a good interface for the `wipe` angle options? You called the option `--transition-pos`, which is weird at first since the options changes the *angle*.

flick0:
ok, what about adding aliases of the previous flags which wud do what they originally did and also have the new flags? imo the new transition names represent the transition better. and also add a `--transition-angle` flag for wipe

and the curves crate is not used, i was messing around with it before i found the keyframes crate and mustv forgot to remove it .
it can be removed without any harm

keyframe crate is used for the bezier interpolation on all transitions except fade/simple so it is being used where ever it can be

also i wont be able to use my pc till the next 2 days, so ill work on this once im back

Horus645:
> ok, what about adding aliases of the previous flags which wud do what they originally did and also have the new flags? imo the new transition names represent the transition better.

I agree with the following being better names:

`outer` -> `shrink`
`simple` -> `fade`
`from` -> `grow`

However, I fell like changing the `bottom`, `top`, `left` and `right` transitions to `wipe`, even though technically more accurate, don't add much in the way of ergonomics. I also think we can leave `any` and `center` as-is. But, I will think a bit more about it.

> keyframe crate is used for the bezier interpolation on all transitions except fade/simple so it is being used where ever it can be

Oh, great. I had only a quick look at it before, my bad for not seeing that. I wonder if we can't emulate its effect manually somehow (shouldn't be that difficult, actually). Not totally necessary, of course, but I generally prefer using as little dependencies as possible. I had a closer look at it now, and it doesn't seem like the code got awfully simpler, though maybe that's just a matter of getting used to it.

> also add a --transition-angle flag for wipe

Yes, this would be better, in my opinion. We would also have to explain well what we mean, though it seems you already documented that.

> also i wont be able to use my pc till the next 2 days, so ill work on this once im back

No problem. There's no rush, take your time :)
I myself will remain very busy for at least one more week.

flick0:
yes, about the beziers, is it fine to use the keyframe lib or maybe extract parts of it into our code, *or make our own*

PR #38, opened 2022-10-03
'wipe' transition

flick0:
something like 'left' but works with a specified angle

### new in flags
+ `--transition-angle` angle in degrees for the 'wipe' transition
+ `--transition-type wipe` new transition option

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -188,15 +188,15 @@ pub struct Img {
     ///
     ///Possible transitions are:
     ///
-    ///simple | left | right | top | bottom | center | outer | any | random | wipe
+    ///simple | left | right | top | bottom | wipe | center | outer | any | random
     ///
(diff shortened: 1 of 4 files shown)


src/cli.rs
```diff
@@ -186,14 +188,16 @@ pub struct Img {
     ///
     ///Possible transitions are:
     ///
-    ///simple | left | right | top | bottom | center | outer | any | random
+    ///simple | left | right | top | bottom | center | outer | any | random | wipe
```
Horus645, line 191: This is an incredibly annoying nitpick, but, since bellow you explain each transition in the order:
`..., center, outer, wipe, any, random`
these should also be in that exact order: `... | center | outer | wipe | any | random`

Horus645 (commented):
This looks really, really good. Sorry it took me so long to get to it. In any case, there's only one very small nitpick, and we should be good to go.

flick0:
made the changes, and np you can take your time

PR #44, opened 2022-10-18
'grow' transition and 'outer' transition with ability to set transition position

flick0:
`grow` transition merges `center` and `any` transitions ( `center` and `any` transitions work as aliases now )
and a new `--transition-pos` flag

### changes
+ new `--transition-pos` flag:
  the position for `grow` and `outer` transitions, accepts coords in format `<x_pos>,<y_pos>` 
  the values for x_pos and y_pos can be given as:
   + percentages (floats between 0 and 1.0) like `0.2,0` and `0.8,0.7` 

    you can also use keywords `'center' | 'top' | 'left' | 'right' | 'bottom' | 'top-left' | 'top-right' | 'bottom-left' | 'bottom-right'` as aliases

    for example:
    ```bash
    --transition-pos 0,0.25     # 0 pixels from left and 25% pixels from bottom
    --transition-pos 0.5,0.5     # 50% pixels from left and 50% pixels from bottom
    --transition-pos top-left  # alias to `0,1`
    ```
   ( this flag will be overridden if user uses `center` and `any` transitions )
+ `--transition-type grow` is like `center` but uses `--transition-pos` flag
+ `--transition-type outer` can also use the `--transition-pos` flag
 

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -75,6 +75,7 @@ pub enum TransitionType {
     Any,
     Random,
     Wipe,
+    Grow,
 }
 
 impl std::str::FromStr for TransitionType {
@@ -87,13 +88,14 @@ impl std::str::FromStr for TransitionType {
             "right" => Ok(Self::Right),
             "top" => Ok(Self::Top),
             "bottom" => Ok(Self::Bottom),
+            "wipe" => Ok(Self::Wipe),
+            "grow" => Ok(Self::Grow),
             "center" => Ok(Self::Center),
             "outer" => Ok(Self::Outer),
             "any" => Ok(Self::Any),
             "random" => Ok(Self::Random),
-            "wipe" => Ok(Self::Wipe),
             _ => Err("unrecognized transition type. Valid transitions are:\
-                     simple | lest | right | top | bottom | center | outer | random | wipe\
+                     simple | lest | right | top | bottom | wipe | grow | center | outer | random\
                      see swww img --help for more details"),
         }
     }
@@ -188,17 +190,21 @@ pub struct Img {
     ///
     ///Possible transitions are:
     ///
-    ///simple | left | right | top | bottom | wipe | center | outer | any | random
+    ///simple | left | right | top | bottom | wipe | grow | center | any | outer | random
     ///
     ///The 'left', 'right', 'top' and 'bottom' options make the transition happen from that
     ///position to its oposite in the screen.
     ///
     ///'wipe' is simillar to 'left' but allows you to specify the angle for transition (with the --transition-angle flag).
-    /// 
-    ///'center' causes a transition from the center to the edges of the screen, while 'outer' is
-    ///the oposite of that.
     ///
-    ///'any' is like 'center', but selects a random spot to start the transition from
+    ///'grow' causes a growing circle to transition across the screen and allows changing the circle's center\
+    /// position (with --transition-pos flag).
+    ///
+    ///'center' an alias to 'grow' with position set to center of screen.
+    ///
+    ///'any' an alias to 'grow' with position set to a random point on screen.
+    ///
+    ///'outer' same as grow but the circle shrinks instead of growing.
     ///
     ///Finally, 'random' will select a transition effect at random
     #[arg(short, long, env = "SWWW_TRANSITION", default_value = "simple")]
@@ -245,6 +251,16 @@ pub struct Img {
     ///Note that the angle is in degrees, where '0' is right to left and '90' is top to bottom, and '270' bottom to top
     #[arg(long, env = "SWWW_TRANSITION_ANGLE", default_value = "0")]
     pub transition_angle: f64,
+
+    ///This is only used for the 'grow','outer' transitions. It controls the center of circle (default is 'center').
+    ///
+    ///position values can be given in the format of '<x_pos>,<y_pos>', for example '700,0' is 700px from left and 0px from bottom
+    ///x_pos and y_pos can also be percentages, which are parsed according to screen size, for example '50%,50%' for center
+    ///
+    ///the value can also be an alias which will set the position accordingly):
+    /// 'center' | 'top' | 'left' | 'right' | 'bottom' | 'top-left' | 'top-right' | 'bottom-left' | 'bottom-right'
+    #[arg(long, env = "SWWW_TRANSITION_POS", default_value = "center")]
+    pub transition_pos: String,
 }
 
 #[cfg(test)]
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -316,36 +316,37 @@ fn parse_coords(raw: &str) -> Result<(f32,f32), String> {
         x = coords[0];
         y = coords[1];
     }
-    let parsed_x: f32;
-    let parsed_y: f32;
+    
+    
 
     //parse x coord
-    match x.parse::<f32>() {
+    let parsed_x = match x.parse::<f32>() {
         Ok(x) => {
-            if x > 1.0 || x < 0.0{
+            if !(0.0..=1.0).contains(&x){
                 return Err(format!(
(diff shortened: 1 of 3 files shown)


ThatOneCalculator:
lgtm

src/cli.rs
```diff
@@ -188,17 +190,21 @@ pub struct Img {
...
     ///The 'left', 'right', 'top' and 'bottom' options make the transition happen from that
     ///position to its oposite in the screen.
     ///
     ///'wipe' is simillar to 'left' but allows you to specify the angle for transition (with the --transition-angle flag).
-    /// 
-    ///'center' causes a transition from the center to the edges of the screen, while 'outer' is
-    ///the oposite of that.
     ///
-    ///'any' is like 'center', but selects a random spot to start the transition from
+    ///'grow' causes a growing circle to transition across the screen and allows changing the circle's center\
```
Horus645, line 200: That '\' after center is unneeded.

src/cli.rs
```diff
@@ -87,13 +88,14 @@ impl std::str::FromStr for TransitionType {
...
+            "wipe" => Ok(Self::Wipe),
+            "grow" => Ok(Self::Grow),
             "center" => Ok(Self::Center),
             "outer" => Ok(Self::Outer),
             "any" => Ok(Self::Any),
             "random" => Ok(Self::Random),
-            "wipe" => Ok(Self::Wipe),
             _ => Err("unrecognized transition type. Valid transitions are:\
-                     simple | lest | right | top | bottom | center | outer | random | wipe\
+                     simple | lest | right | top | bottom | wipe | grow | center | outer | random\
```
Horus645, line 98: lest -> left (this probably was wrong before, but we might as well fix it here)

src/daemon/processor/animations.rs
```diff
@@ -99,9 +103,6 @@ impl Transition {
             2 => self.right(new_img, outputs, sender, stop_recv),
             3 => self.top(new_img, outputs, sender, stop_recv),
             4 => self.bottom(new_img, outputs, sender, stop_recv),
```
Horus645, line 105: We are missing 'grow' and 'outer' transitions in here. For 'grow' we can simply call 'any'. For 'outer' we might be ok with leaving it out for now. Nevertheless, **remember to update the random value that comes before this**. Currently it is `match r % 8`, which will generate numbers in the range `[0-7]`, but we only have numbers in the range `[0-4]` in this PR.

src/daemon/processor/mod.rs
```diff
@@ -43,21 +43,39 @@ pub struct ProcessorRequest {
...
+                    rand::random::<u32>() % dimensions.1,
+                );
+                TransitionType::Grow
+            }
+            _ => img.transition_type.clone(),
+        };
         Self {
             outputs: vec![info.name.to_string()],
-            dimensions: info.real_dim(),
+            dimensions: dimensions,
```
Horus645, line 69: This can be just `dimensions,`.

src/daemon/processor/mod.rs
```diff
@@ -43,21 +43,39 @@ pub struct ProcessorRequest {
...
             old_img,
             path: img.path.clone(),
-            transition_type: img.transition_type.clone(),
+            transition_type: transition_type,
```
Horus645, line 72: Same: `transition_type,`.

src/daemon/processor/mod.rs
```diff
@@ -43,21 +43,39 @@ pub struct ProcessorRequest {
...
             speed: img.transition_speed,
             filter: img.filter.get_image_filter(),
             step: img.transition_step,
             fps: Duration::from_nanos(1_000_000_000 / img.transition_fps as u64),
             angle: img.transition_angle,
+            pos: pos,
```
Horus645, line 78: Same.

src/cli.rs
```diff
@@ -245,6 +251,16 @@ pub struct Img {
...
+
+    ///This is only used for the 'grow','outer' transitions. It controls the center of circle (default is 'center').
+    ///
+    ///position values can be given in the format of '<x_pos>,<y_pos>', for example '700,0' is 700px from left and 0px from bottom
+    ///x_pos and y_pos can also be percentages, which are parsed according to screen size, for example '50%,50%' for center
+    ///
+    ///the value can also be an alias which will set the position accordingly):
+    /// 'center' | 'top' | 'left' | 'right' | 'bottom' | 'top-left' | 'top-right' | 'bottom-left' | 'bottom-right'
+    #[arg(long, env = "SWWW_TRANSITION_POS", default_value = "center")]
+    pub transition_pos: String,
```
Horus645, line 263: This should be a (f32, f32). Look all the way down for more info.

src/daemon/processor/mod.rs
```diff
@@ -309,3 +328,102 @@ fn send_frame(
...
+                }
+                parsed_y = parsed_y * dimensions.1 / 100;
+            } else {
+                debug!("Invalid y coord, using 0 as fallback: {}", y);
+                parsed_y = 0;
+            }
+        }
+    }
+    (parsed_x, parsed_y)
+}
```
Horus645, line 429: All of this should be done when we parse the **command line arguments**, in `cli.rs`, not here. If we detect something invalid, we immediately exit, tell the user and don't even send anything to the processor.

Here's how you can do that:

First, transform `transition_pos` into a `(f32, f32)`. Each of those is a number between 0 and 1 that represents a percentage. You want to parse these values from the command line. So you will call this parsing function to evaluate the command line arguments. You do this like so:

```rust
#[arg(long, env = "SWWW_TRANSITION_POS", value_parser = parse_coords, default_value = "center")]
pub transition_pos: (f32, f32),
```

And then your `parse_coords` will have to return a `Result<(f32,f32), String>`. All of those `debug` statements at incorrect locations will transform into `Err(format!(msg))`.

After all this, if the parsing succeds we will have the correct numbers to send to the daemon, and we won't have to check for any errors. All we have to do is to transform the percentages into screen coordinates.

Now, there is one problem: this only allows for the percentage approach you took, not the absolute location of the pixels. Personally, I think that's fine; I believe it makes more sense for this option to be a percentage than an absolute value. We can make it more complicated later if needed be.

Horus645 (changes requested):
Overall, I like it a lot. The hard part (implementing the `grow` logic) is flawless, as far as I can tell.

The biggest thing is that `parsing` should really only be done when reading the command line arguments. That is, we must make sure the command line arguments are correctly formatted **before** sending them to the daemon, not after. Overall, this shouldn't be too difficult (just move some stuff around and rework some minor stuff, really). The solution I proposed bellow has the unfortunate consequence of not allowing people to use absolute values, but I believe that is a fair compromise.

src/daemon/processor/animations.rs
flick0: how about the `any` transition being both grow and outer randomly?

Horus645: That sounds pretty good!

src/cli.rs
```diff
@@ -245,8 +251,105 @@ pub struct Img {
...
+                return Err(format!(
+                    "y coord not in range [0,1.0]: {}",
+                    y
+                ))
+            }
+            parsed_y = y
+        },
+        Err(_) => return Err(format!("Invalid y coord: {}", y)),
+    }
+    Ok((parsed_x, parsed_y))
```
Horus645, line 349: This can be:

```rust
let parsed_x = match x.parse::<f32> {
    ...
};

let parsed_y = match y.parse::<f32> {
    ...
};
```

(that is, you don't have to declare them in advance)

Horus645 (changes requested):
Just some style stuff left!

I recommend running `cargo clippy` before sending in your PRs, as that shows us all the un-idiomatic stuff we wrote. (Nota that, for this PR, clippy will give us a warning about a function having too many arguments; you can ignore that for now)

PR #50, opened 2022-11-15
Bezier curves

### LINKED ISSUE #30: `--transition-duration` to control transition duration
### Description

Currently the speed of the transition must be configured via `--transition-step` and `--transition-speed`. Tweaking these values require some pretty delicate understanding of the how transitions are implemented, keeping in mind screen resolution, and possibly some basic math.

I believe the current approach is designed this way since it directly exposes the internals of  the transitions implementations.

### Possible improvement

I'd like to propose merging these into a `--transition-duration` flag, which merely takes an integer with the transition duration in milliseconds. This is something a lot "friendlier" and handles just one value that's very simple to understand.

The correct values for `transition-speed` and `transition-step` can be calculated based on `--transition-duration` and `--transition-fps`.

flick0:
bezier curve interpolation for all transitions(except `simple`), basically ability to hv them ease-in/ease-out etc...

### changes
+ added default ease-in-out interpolation to all transitions

+ new `--transition-bezier` flag:
  >  takes 4 float values separated by commas(https://cubic-bezier.com is a good website to get them from)
     defaults to `.54,0,.34,.99` (close to ease-in-out)

+ new `--transition-duration` flag, replaces `--transition-speed`
  > duration in seconds(float)

also resolves #30


flick0:
the transition is a bit laggy for some reason rn

src/daemon/processor/animations.rs
```diff
@@ -112,6 +117,13 @@ impl Transition {
         }
     }
 
+    fn bezier_seq(&self,start:f32,end:f32) -> (AnimationSequence<f32>,Instant){
+        (keyframes![
+            (start,0.0,self.bezier),
+            (end,self.duration as f32, self.bezier)
```
WhyNotHugo, line 123: `self.duration` is already f32.

src/cli.rs
```diff
@@ -223,19 +223,11 @@ pub struct Img {
...
-    ///controls how fast the 'going to the right' motion goes.
-    ///
-    ///A value of 'n' means that we'll go 'n' columns at a time for the 'left' transition, for
-    ///example. For the transitions that work with radii ('center', 'outer' and 'any') a value of
-    ///'n' means that we'll increase the radius by 'n' every time.
-    #[arg(long, env = "SWWW_TRANSITION_SPEED", default_value = "5")]
-    pub transition_speed: u8,
+    ///Note that this doesnt work with the 'simple' transition
+    #[arg(long, env = "SWWW_TRANSITION_DURATION", default_value = "3")]
+    pub transition_duration: f32,
```
WhyNotHugo, line 230: Specify the unit in the description (e.g.: `in seconds`).

WhyNotHugo:
I believe this can apply to the `simple` transition too.

Currently one image is "blended" into another at a linear rate. This could be done applying a curve (e.g.: like the one on this PR). That can be done separately though, I don't see any issues here.

Tested this PR with the `outer` curve and works fine for me, having the right curve makes a big difference!

src/daemon/processor/animations.rs
flick0: ah,missed that

src/cli.rs
flick0: i did add it, its on the line right above this snippet

flick0:
> I believe this can apply to the `simple` transition too.
> 
> Currently one image is "blended" into another at a linear rate. This could be done applying a curve (e.g.: like the one on this PR). That can be done separately though, I don't see any issues here.
> 
> Tested this PR with the `outer` curve and works fine for me, having the right curve makes a big difference!

idt non-linear animation to fade would really work, iirc simple uses transition-step for the speed of fade, so maybe can use that with bezier curves

Horus645:
I am calling this with:
```bash
./target/release/swww img --transition-type=left --transition-bezier 1,1,1,0.5 ~/Pictures/Backgrounds/Aria.jpg
```
And I am consistently getting:

image

It stops at this point for some reason. Note that for different values it still stops before going through the full screen, it just stops closer to the edge.

Another thing: calling the curve with `0,0,0,0`. Makes `swww` enter an infinite loop, without ever changing the background.

Also, from my testing, there doesn't seem to be a way to control overall speed of the transition? I'd like to be able to do that. So, if we chose to use a linear transition, for example, we can still control how fast we sweep through the screen.

flick0:
+ it stops and then immediately completes the transition as the curve itself is like that
image
it has an instant jump near the end

+ hmm 0,0,0,0 shud just be a linear curve, but we can add a check for it (0,0,1,1) is also linear and ppl can use that instead

+ the speed should be controllable through beziers or by changing transition duration but im not sure how we can implement a speed flag again with beziers yet

Horus645:
> it stops and then immediately completes the transition as the curve itself is like that

That's what I am saying, I've tested it and it actually never completes. That dark spot stays there forever.

> the speed should be controllable through beziers or by changing transition duration but im not sure how we can implement a speed flag again with beziers yet

In that case, I think we should change the flags name. `duration` is meant to be a simple flag, where we calculate all the necessary values to give you a transition that lasts roughly what you asked for. It's actually a really annoying problem, since we have so many different variables that affect how long a transition will last. You don't have to implement it in this PR.

flick0:
> That's what I am saying, I've tested it and it actually never completes. That dark spot stays there forever.
hmm, when i tried on my machine it finishes the transition, maybe because the transition is happening too fast?

>It's actually a really annoying problem, since we have so many different variables that affect how long a transition will last.
can u giv an example? ive tried taking into account as many vars as i can but i might be missing something

>  You don't have to implement it in this PR.
the duration flag actually helps alot in adding beziers, and it makes the overall transition smoother (reduced artifacts) and more reliable as the transition progress doesn't depend on number of frames it took anymore, it depends on time since transition started

flick0:
also about the darkspot, can you try the wipe transition equivalent for that?

Horus645:
This last change you pushed seems to have fixed it.

> what do u think about aliasing left and right to wipe?

I feel like this would be pretty good regardless. In fact, you can alias everything: `left`, `right`, `top` and `bottom`. However, I was thinking of doing this in a separate PR at a later date. If you think it's not too much trouble, you can try doing it now.

> the duration flag actually helps alot in adding beziers, and it makes the overall transition smoother (reduced artifacts) and more reliable as the transition progress doesn't depend on number of frames it took anymore, it depends on time since transition started

I see. In that case, we should keep it.

src/daemon/processor/animations.rs
```diff
@@ -336,43 +268,49 @@ impl Transition {
...
-                    let diff_y = pix_y.abs_diff(center_y as usize);
-                    let pix_center_dist = diff_x * diff_x + diff_y * diff_y;
-                    if pix_center_dist >= dist_center * dist_center {
+                    let diff_x = pix_x.abs_diff(center_x as usize) as f32;
+                    let diff_y = pix_y.abs_diff(center_y as usize) as f32;
+                    let pix_center_dist = diff_x.powf(2.0) + diff_y.powf(2.0);
+                    if pix_center_dist >= dist_center.powf(2.0) {
                         let step =
-                            self.step.saturating_add((pix_center_dist - (dist_center * dist_center)) as u8);
+                            self.step.saturating_add((pix_center_dist - dist_center.powf(2.0)) as u8);
```
Horus645, line 299: Same, could you please change it to:

```rust
let pix_center_dist = (diff_x.powf(2.0) + diff_y.powf(2.0)).sqrt();
       if pix_center_dist >= dist_center {
              let step = self
                     .step
                      .saturating_add((pix_center_dist - dist_center).log2() as u8);
```

Horus645 (commented):
Alright, I am almost ready to merge this! Just two really minor things. I noticed that the `outer` and `grow` transition are changing the color too fast. So, could you please just change the way we are calculating the `step` for them? I've put the new versions in the review. This makes the transition smoother when the step is very low.

Other than that, just run `rust fmt` on the files, squash the commits in this PR, and we should be done!

PR #53, opened 2022-12-01
add new transition previews

flick0:
adds these transition gifs to readme:

https://i.imgur.com/NAWKiEW.gif
https://i.imgur.com/ycrTPBP.gif
https://i.imgur.com/1w34rfW.gif

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 ### Efficient animated wallpaper daemon for wayland, controlled at runtime
 
 ![animated gif demonstration](https://i.imgur.com/Leuh6wm.gif)
-![image transition demonstration](https://i.imgur.com/NAWKiEW.gif)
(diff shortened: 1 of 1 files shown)


Horus645:
Nice! I was just thinking our previews were woefully out of date. 

Originally I uploaded my previews to imgur and just referenced them here, because I don't want people to have to download big `gif` files when they just want the binaries.

Now, I have found a better solution: create a specific branch just to host these assets, and then reference them like so:
`outer transition demonstration`

So, instead of `./assets/...` we would have `../demos/assets/...`.

I've just created the `demos` branch. What this means is that the `gif` themselves need to be merged into the `demos` branch (so, we'll need another PR), while this PR just needs to reference them, as demonstrated above.

PR #54, opened 2022-12-01
Tweak README slightly

aouerfelli:
Minor fixes and polish to keep things consistent.


Horus645:
Ah, many thanks. I had to do this sooner or latter but just kept on delaying it.

PR #63, opened 2022-12-11
change default transition step value

flick0:
i feel like rn default transition step is more on the "too smooth" side,

this pr changes it to `90` and it shud be on a nice middle ground.


Horus645:
Agreed.

PR #66, opened 2022-12-13
new transition: wave

flick0:
# changes
+ adds a new cli flag `transition-scale` to control size of each single wave
  > value is given in format `width,height`

+ adds new transition `wave`, its like wipe but more wavy

wave

> the graph/formula used here can also work with any given function ( im using `f(x) = sin(x)` for the waves rn ) so i was thinking about a `math` transition that lets users input their own functions, what do you think?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,7 +25,6 @@ image = "0.24"
 fast_image_resize = "2.3"
 clap = { version = "4.0", features = ["derive", "wrap_help", "env"] }
 rand = "0.8"
-meval = "0.2"
 utils = { path = "utils" }
 
 [dev-dependencies]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,7 +25,7 @@ image = "0.24"
 fast_image_resize = "2.3"
 clap = { version = "4.0", features = ["derive", "wrap_help", "env"] }
 rand = "0.8"
-
+meval = "0.2"
 utils = { path = "utils" }
 
 [dev-dependencies]
--- a/daemon/src/processor/animations.rs
+++ b/daemon/src/processor/animations.rs
@@ -37,7 +37,7 @@ pub struct Transition {
     angle: f64,
     pos: (f32, f32),
     bezier: BezierCurve,
-    scale: (f32, f32),
+    wave: (f32, f32),
 }
 
 /// All transitions return whether or not they completed
@@ -66,7 +66,7 @@ impl Transition {
                     y: transition.bezier.3,
                 },
             ),
-            scale: transition.scale,
+            wave: transition.wave,
         }
     }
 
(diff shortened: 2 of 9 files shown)


Horus645:
Hmmm.

No, I don't think we should allow for arbitrary math functions. Primarily because I believe only periodic functions (line `sin`) make any sense. If you could show me something interesting that you achieved with a different function, I'd be willing to consider it.

`transition-scale` is a terrible name, in my opinion. I believe `transition-wave` would be better.

Finally, if we do this, then we should drop the `wipe` transition in the daemon, and transform it into an alias to `wave`, with `--transition-wave 0,0`.

flick0:
> Finally, if we do this, then we should drop the `wipe` transition in the daemon, and transform it into an alias to `wave`, with `--transition-wave 0,0`.

I thought we cud have both, because the `wave` transition is way heavier than the standard `wipe`

Cargo.toml
```diff
@@ -25,7 +25,7 @@ image = "0.24"
 fast_image_resize = "2.3"
 clap = { version = "4.0", features = ["derive", "wrap_help", "env"] }
 rand = "0.8"
-
+meval = "0.2"
```
Horus645, line 28: I believe we don't really use this anywhere?

daemon/src/processor/animations.rs
```diff
@@ -344,12 +425,7 @@ mod tests {
     #[test]
     fn transitions_should_end_with_equal_vectors() {
         use TransitionType as TT;
-        let transitions = [
-            TT::Simple,
-            TT::Wipe,
-            TT::Outer,
-            TT::Grow,
-        ];
+        let transitions = [TT::Simple, TT::Wipe, TT::Outer, TT::Grow];
```
Horus645, line 428: Could you please add a test case for our new wave transition?

src/cli.rs
```diff
@@ -92,9 +93,10 @@ impl std::str::FromStr for TransitionType {
             "center" => Ok(Self::Center),
             "outer" => Ok(Self::Outer),
             "any" => Ok(Self::Any),
+            "wave" => Ok(Self::Wave),
             "random" => Ok(Self::Random),
             _ => Err("unrecognized transition type.\nValid transitions are:\n\
-                     \tsimple | left | right | top | bottom | wipe | grow | center | outer | random\n\
+                     \tsimple | left | right | top | bottom | wipe | grow | center | outer | random | math\n\
```
Horus645, line 99: `math` -> `wave`

Horus645 (commented):
Looks good! Only some really minor stuff.

Horus645:
Sorry it took me a while. End of the year and stuff.

Horus645:
Looks pretty good! Took me a while, my bad.

PR #71, opened 2023-01-08
add ability to use screen cords instead of percentages in `--transition-pos`

flick0:
if `--transition-pos` is not a float it will interpret it as screen position

ie:
 +  `0.1,0.8` will be parsed as percentages
 +   `700,200` will be parsed as pixels
 + ~~it wil also raise an error if its `700.0,200.0` since it tries to parse that as percentage~~

<hr>

https://github.com/Horus645/swww/pull/71/commits/2d4c1a09d8b151c3b7e1b57db9c62ffab0a7982f

+ added a new `Position` enum
+ no longer raises an error when float values/percentages are not in range 0..1 rather it prints a warning about it
  >why?, because we dont check if position is out of screen with pixels too and it might useful to hv position set out of screen as the animation changes a bit
+ ~~you can no longer send values like `100,0.7` (a mix of percentage and pixel) due to how `Position` enum works~~

this allows stuff like starting transition from cursor pos:
```
swww img img.png --transition-type grow --transition-pos "$(hyprctl cursorpos)"
```


Horus645:
I like this idea, but I think we should just always parse it as an `f32`. Trying to parse something like `1` (without the trailing decimals) as a float **does** work. So I think instead of storing those bools we can just do the following:

```rust
if coord > 0 && coord < 1.0 {
      // Treat it as a percentage
} else {
      // Treat it as a pixel position
      // In this case, we have to make sure to call round() before transforming the float into u32
      // AND, we need to have to make sure the value does not exceed the screens' dimensions.
}
```

Like this, we wouldn't have to awkardly store those extra `bool` variables in `transition_pos`.

flick0:
> Trying to parse something like 1 (without the trailing decimals) as a float does work

it never tries parsing as float if it doesnt find a decimal point, so that kind of situation shudnt be possible

as for the awkward `bool` value, i agree, maybe we can hv a separate Position struct for that

flick0:
or even totally different flags for both percentage values and pixel values

Horus645 (commented):
Overall, I think just using an Enum for the values in `transition_pos` would improve overall readability a lot. See what you think.

flick0:
+ added a new `Position` enum
+ no longer raises an error when float values/percentages are not in range 0..1 rather it prints a warning about it
  >why?, because we dont check if position is out of screen with pixels too and it might useful to hv position set out of screen as the animation changes a bit
+ you can no longer send values like `100,0.7` (a mix of percentage and pixel) due to how `Position` enum works

utils/src/communication.rs
```diff
@@ -9,6 +9,28 @@ use std::{
 
 use crate::comp_decomp::BitPack;
 
+#[derive(PartialEq, Clone, Serialize, Deserialize, Debug)]
+pub enum Position {
+    Pixel(f32, f32),
+    Percent(f32, f32),
+}
```
Horus645, line 16: Why don't you make this store just a singular value? (just one `f32`). Then you could just send a tuple of `Position` and you could store a mix of percentage and absolute values

Horus645 (commented):
> you can no longer send values like 100,0.7 (a mix of percentage and pixel) due to how Position enum works

I think we can allow for this pretty easily. Just make the `Position` enums store only value, and use a `(Position, Position)` tuple.

flick0:
added a `Cord` enum and made `Position` a struct of `Cord`s

PR #79, opened 2023-02-03
fix reading image from stdin (don't cannonicalize)

### LINKED ISSUE #42: Supporting reading the image from stdin
This is a weird request, but I wish it's possible to do something like this
`cat wallpaper.png | swww img -`
where `-` stands for read from standard input as most program do

Horus645:
We can't cannonicalize the '-' path, so let's not do that. Also added a regression test to avoid this happenning again.

Fixes #42 (again)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 Fixes:
   * `swww query` not returing the image being displayed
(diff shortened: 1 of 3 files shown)


Horus645:
whoops. Fixed in the next commit

PR #80, opened 2023-02-04
[WIP] cache the images to reload them on startup

### LINKED ISSUE #65: daemon unsets on screen shutdown
I'm Not sure if it's intended behavior, so i'll refrain from assigning a label.

Upon physically powering off a screen, swww doesn't kill the daemon but unsets itself. On screen powerup i need to manually use the swww img command to set the image again. 

Is it possible to store the loaded image and when swww detects a screen powering on (and thus swww query returns the screen data) to autoload what was previously on that screen?

Horus645:
This creates a directory in `$XDG_CACHE_HOME/swww` or `$HOME/.cache/swww` to store the images currently being displayed. We then load them up when a new bg shows up in the daemon.

Fixes #65.

IMPORTANT: this is currently WIP. It works, but the code is ugly, full of `unwraps` and repeated stuff. I need to clean this up and possibly add a few test cases to make sure nothing explodes.


Horus645:
This should be working now, if anyone would like to try using it.

I will take it for a ride myself for a few days before merging it.

PR #82, opened 2023-02-08
Sync outputs

### LINKED ISSUE #39: Span image over all monitors.
Would it be possible to allow an image to be spanned over multiple monitors?
I could try to implement this myself. I would just need a hint, where in the code the image is drawn on a specific output.
I gues I would just need to split the image data into `n` chunks and then send them to the appropriate outputs.

Horus645:
Fixes #39.

Sorry it took so long, @FF-AntiK. This turned out to be a lot more annoying than I expected. Hopefully this solution here is good enough.

### DIFF
--- a/daemon/src/main.rs
+++ b/daemon/src/main.rs
@@ -433,6 +433,7 @@ fn main_loop(
             while i != bgs.len() {
                 if let Some(should_remove) = bgs[i].handle_events() {
                     let mut processor = processor.borrow_mut();
+                    processor.set_output_count(bgs.len() as u8);
                     processor.stop_animations(&[bgs[i].info.name.clone()]);
                     if should_remove {
                         bgs.remove(i);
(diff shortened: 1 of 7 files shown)


Horus645:
Oops, I tried running this on my notebook and discovered that I forgot to `git add` a file :facepalm:

FF-AntiK:
Hi @Horus645 ,
thank you very much for working on this.
I've tried this branch, but it gives me the error when sending an image to any monitor (both with and without `--sync`):
`Error: "Failed to receive answer: io error: Resource temporarily unavailable (os error 11)"`

Horus645:
@FF-AntiK did you by any chance test it by rnning `./target/release/swww init`? Note in that `swww` will use the `swww-daemon` binary in your `$PATH`. Since we changed the daemon in this PR, in order to test it, you'll have to start the daemon manually with `./target/release/swww-daemon`, and *then* use `./target/release/swww img` to set the wallpapers. Annoying, I know, but I am not sure what to do about this.

FF-AntiK:
Oh, it's possible, that I've only built `swww` and didn't update `swww-daemon`. As soon as I have time to test it agein I'll let you know :-)

FF-AntiK:
I've tested it now and it works great!
Thanks again for your excelent work! :-)

PR #88, opened 2023-02-19
fix: Missing `/` when using `$HOME/.cache/swww`

max-ishere:
Replaced error prone `env + "path"` with `PathBuf.push('path')` Previously, cache dir would not be avaliable if `$XDG_CACHE_HOME` was not defined


Horus645:
LGTM. Thanks for this!

max-ishere:
You're welcome.

PR #93, opened 2023-02-25
wait until all gif's frames are ready

### LINKED ISSUE #89: Animated gif hangs temporarily on first frame
Pointed out by @max-ishere in #86, because the client sends the image before the full gif processing is complete, `swww` will sometimes "hang" in the first frame, which can be jarring in some scenarios.

Horus645:
Fixes #89.

@max-ishere could you please test this to see if it's good enough? Ultimately, the problem is that sending many animated frames through the socket takes time. So, if the transition is very fast, and there are many frames to send, right now there is pretty much nothing we can do (we'd have to change the way to send the images through the socket, which is a much more annoying rewrite than this).

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -8,7 +8,6 @@ use std::{
     os::unix::net::UnixStream,
     path::{Path, PathBuf},
     process::Stdio,
-    thread::JoinHandle,
     time::Duration,
 };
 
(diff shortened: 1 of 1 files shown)


max-ishere:
What it does now is not show the first frame until its all ready to go? I'm not familiar with the code base so cant evaluate the diffs.

Looks good to me. In fact I think this is exactly the thing i needed.

I noticed is that if the transition duration to too long the GIF wont play until the transition has ended. On lower duration its not as noticeable.

max-ishere:
As far as hanging frames - they are gone now. Still takes a bit to load. Maybe add an option to export cached files from .cache/swww and use them instead of gif? but idk if that is aldready too complex for a minimalistic swww

max-ishere:
[README.md ##Resizing GIFs](https://github.com/Horus645/swww#high-cpu-usage-during-caching-of-a-gifs-frames)

Probably the reason for slow loading in my case. The gif I use is low res, Ill try to find a better source video and use that instead. Not a priority for me right now tho.

Horus645:
> I noticed is that if the transition duration to too long the GIF wont play until the transition has ended. On lower duration its not as noticeable.

This is intentional. Playing the gif while the transition is on-going would be very complex.

> Still takes a bit to load. Maybe add an option to export cached files from .cache/swww and use them instead of gif?

Oh no, the cache is not meant to load the images faster; it's meant to allow for restoring the last used image when rebooting your pc. The cache itself is egregiously simple. For example, it stores the images as raw RBGA vectors, which is very space-inefficient, but very convenient for us. We *do* compress the animation frames, but that's basically it.

> The gif I use is low res, Ill try to find a better source video and use that instead.

You might not have to do that. You can try different filters instead. `--filter Nearest` or `--filter Bilinear` should both be faster than the default.

max-ishere:
I found a GIF bug, cannot describe it yet ill post after I do some testing

max-ishere:
maybe its alright i am trying to figure out how to reproduce it and what GIFs cause it and stuff like that

Horus645:
> maybe its alright i am trying to figure out how to reproduce it and what GIFs cause it and stuff like that

If just have an example `gif` to send I could try helping. Note that, if it isn't related to this thread, I'd prefer you opened a new issue.

max-ishere:
I am not sure what is causing this and how to reproduce this. I had this exact issue in the previous version as well. The symptoms are that the GIFs don't play. But the more consistent indicator that I remember having is that `clear` cycles through GIF frames and never clears: The GIF plays for less than a second and stops on some frame and if you call `clear` again it does the same thing.

Restarting the daemon fixes the issue: the GIF starts to play. I had this issue only while testing my script and I assume there is some race condition that I caused because I called the script too often. Almost never did I have such issues under normal operation: GIF on login and `.png` afterwards + sometimes executing that behavior manually, not just on login.

I assume because I was testing `swww` with different timings and GIFs and different verions and all that I messed up the daemon or something.

max-ishere:
> Oh no, the cache is not meant to load the images faster; it's meant to allow for restoring the last used image when rebooting your pc.

Will this let me play the GIF faster when logging in?
```
cp cached-eDP-1 .cache/swww/eDP-1
swww init
sleep 1
swww img static.png
```
From your explanation I would assume so.

max-ishere:
> If you just have an example `gif` to send I could try helping. Note that, if it isn't related to this thread, I'd prefer you opened a new issue.

It was related because it was a bug I didnt recognize I had, but its a separate issue now as described here: https://github.com/Horus645/swww/pull/93#issuecomment-1445378823

Horus645:
Alright, seems everything's fine on this PR, at least, so I'll be merging.

PR #94, opened 2023-02-25
new transition: `fade`

flick0:
honestly, i thought it wouldn't make much difference to use beziers but after i made a small change and tried it, it made the fade look much much cleaner and polished  (no weird color shifting that usually happens)

added new transition `fade`

https://user-images.githubusercontent.com/77581181/221372512-2710e660-9cf3-4cf9-929e-d321d4f1b626.mp4

### DIFF
--- a/daemon/src/processor/animations.rs
+++ b/daemon/src/processor/animations.rs
@@ -148,7 +148,6 @@ impl Transition {
         self.fallback(new_img, outputs, sender, stop_recv)
     }
 
-
     fn wave(
         mut self,
         new_img: &[u8],
--- a/daemon/src/processor/animations.rs
+++ b/daemon/src/processor/animations.rs
@@ -97,7 +97,7 @@ impl Transition {
         )
     }
 
-    fn simple(
+    fn fallback(
         mut self,
         new_img: &[u8],
         outputs: &mut Vec<String>,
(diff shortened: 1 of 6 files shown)


Horus645:
Sorry, @flick0, I don't like this because now `simple` is not longer... well, simple. I'd be willing to add this maybe as a new type of transition, maybe `fade_in` or something.

flick0:
i was also thinking of a way to totally replace the transition step and instead do it with beziers, maybe a “step-bezier“, or atleast make it a bit more predictable between transitions since it changes drastically from transition to transition, we cud probably do this by handling color changes right inside each transition using time instead of frame count

and yes the actual `simple` transition is still in there but just renamed to fallback since every transition still uses it if it fails, we cud rename it back to simple and rename the new transition to `fade`

i cant work on them right now tho, since i hv exams going on but will make the changes as soon as i can

Horus645:
> i cant work on them right now tho, since i hv exams going on but will make the changes as soon as i can

Take your time. I am kind of busy right now too.

flick0:
> i was also thinking of a way to totally replace the transition step and instead do it with beziers, maybe a “step-bezier“

i was trying this and got it to work on `grow` but the perfomance hit is big


https://user-images.githubusercontent.com/77581181/224919921-44de2601-9faa-49f6-bb02-ae469d6c656b.mp4

i might make a separate pr for it later when i somehow optimize it a bit, for this pr i will only keep the new `fade` transition which uses beziers instead of `step`

Horus645:
I am actually thinking of moving all of the processing logic to the client (including transitions), and then only sending the finished frames to the daemon. This might allow us to compute the transitions faster. But, it is still a ways off, since I am really busy atm, and will be for a least give or take 4 months.

Horus645:
Hey @flick0, is this like, ready to be reviewed/merged? Or are you still planning on changing stuff?

flick0:
yep this pr shud be ready to be reviewed, i have multiple exams for a couple months so i cant work much on the transition step to beziers pr

flick0:
also, do u have discord, id like to add you

Horus645:
> also, do u have discord, id like to add you

Horus#2902. I am not on all that often, though I do usually check things out every Saturday.

flick0:
hey @Horus645 this pr is done btw so do check it out

PR #96, opened 2023-02-25
Add github action to build binaries on tags

HakierGrzonzo:
# Description:

This PR adds a github action, that builds binaries and generates manpages. 

If it is run on a tag, then it creates a release which has a zip file with all the stuff a binary package should need.

An example release is available [here](https://github.com/HakierGrzonzo/swww/releases/tag/v0.1.3). Check if I missed something that should be included. 

## TODO:

- Add hashes to the release body
- GPG signatures?

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,49 @@
+name: Build binaries
+
+on: [push]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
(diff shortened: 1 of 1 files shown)


Horus645:
Don't we also need to ensure `wlr-roots` is installed too?

HakierGrzonzo:
Abandoning this idea

PR #120, opened 2023-05-19
Cache gif animations

Akida31:
swww caches images which takes a really long time for large gifs.
A 160 MB gif takes around 3 minutes to compress.
Taking that long and eating a complete resource is unnecessary because
most switched wallpapers are used more than once.
This commit caches the compressed frames in the `.cache` directory
so that next time the frames can be reused without compression again
(it takes around 3s to load the cache).

Note that the cache is somewhat conservative, if you change some
properties of the animation of the gif it will likely recompress. 
This could be improved in the future but I wanted to be sure that no
outdated animations are displayed.

I don't know if this is a too big feature but for me it is really helpful and I didn't know a way to implement this in a shell wrapper.

If you have any improvements/ suggestions, please say so!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1292,7 +1292,6 @@ dependencies = [
  "log",
  "rand",
  "sd-notify",
- "simplelog",
  "smithay-client-toolkit",
  "utils",
 ]
@@ -1442,6 +1441,7 @@ dependencies = [
  "lzzzz",
  "rand",
  "serde",
+ "simplelog",
 ]
 
 [[package]]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1276,11 +1276,14 @@ name = "swww"
 version = "0.7.3"
 dependencies = [
  "assert_cmd",
+ "bincode",
  "clap 4.2.1",
  "clap_complete",
  "fast_image_resize",
  "image",
+ "log",
  "rand",
+ "serde",
  "utils",
 ]
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,10 +22,13 @@ debug = 1
 strip = false
 
 [dependencies]
+bincode = "1.3"
 image = "0.24"
 fast_image_resize = "2.7"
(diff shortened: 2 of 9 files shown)


Horus645:
Oh man, this looks like it was a lot of work, I am sorry to say this, but if you look at the `sctk-0.17` branch, you'll see I've basically rewritten the entire `daemon`. We actually do not have a decent caching method for that branch yet (we literally just call the client as a separate process).

Because of this, I am afraid this PR will just be kind of useless in the very near future. I have only not merged `sctk-0.17` because I need to have like a whole day to truly test it in a wide range of scenarios and make sure it won't break, at least in my setups (and also update some documentation).

That said, looking at this did give me some ideas we can try using when I get around to implementing proper caching for `sctk-0.17`.

As an aside, you can try testing the `sctk-0.17` branch if you want. It should load and process the gif a little faster, but it will still eat away some processing power. Once it is loaded, however, you should see a dramatic reduction in memory usage.

Akida31:
No problem, this PR was far from useless for me. Even if this will be implemented in a better way in the future, I can use this now :) 
I've tested the caching with a couple of images and can say that it is quite large. Here are some stats:
* 400 MB video files (mp4)
* 2 GB gif files (from gifsicle)
* 6 GB cache files
From that I would say that it could be really nice if swww supported mp4 or a video format like that and it could be helpful to improve the caching to reduce the file size.

If you want I can rebase my changes onto `sctk-0.17`, it looks the parts I changed are quite similar.

Feel free to close this PR if you want or we can keep it open until the rewritten caching is supported so that other people can use this.

Akida31:
I've tried out the "new" version and it looks really nice overall. Also the caching looks good and works for me, thank you for implementing this!
One caveat: I have big/ many animations and so the cache may become quite large. I know how to clean it up but I'm not sure every user does. I think adding a system to automatically clean the cache if it is getting really large or if the version changes (and the old cache isn't used anymore) or adding a subcommand to clear the cache would be valuable.

PR #125, opened 2023-06-26
Fix image padding & add resize-to-fit option

SignalWalker:
1. Currently, images padded from odd widths are skewed. The first commit fixes that.
2. The second commit adds an option to resize images to the borders of the screen, without cropping.

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -1,6 +1,6 @@
 /// Note: this file only has basic declarations and some definitions in order to be possible to
 /// import it in the build script, to automate shell completion
-use clap::Parser;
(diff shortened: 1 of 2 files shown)


Horus645:
Looks pretty good. Could you please also change the documentation in the [`doc`](https://github.com/Horus645/swww/tree/main/doc) to  reflect the changes?

Horus645:
Since this is the only thing left I want to merge before merging #123, I will just accept it as-is and write the documentation myself. Hope you don't mind.

SignalWalker:
Yeah, that's fine; my laptop's motherboard fried itself, so I didn't see that till today.

PR #129, opened 2023-07-01
add flag to invert the y value from --transition-pos

### LINKED ISSUE #127: "--transition-type grow --transition-pos" not working correctly
OS: Arch Linux
WM: Hyprland 0.26.0
swww 0.7.3

**_swww img <path/to/image> --transition-type grow --transition-pos "$(hyprctl cursorpos)" --transition-fps 60 --transition-duration 3_**

This should in theory follow the mouse cursor and start it's animation from there but it is inverted in y-axis.
For eg: If my cursor position is in upper half of monitor the swww wallpaper changing animations starts from just opposite side i.e. lower half of monitor but x-axis is same.

I have watched videos of rices with Hyprland before but swww was working as expected but when I am using it, it's not working the same way.

Here is the script that I run with a keybinding to change the wallpaper at will:-

#!/bin/bash

DIR=$HOME/Pictures/Walls/
PICS=($(ls ${DIR}))

RANDOMPICS=${PICS[ $RANDOM % ${#PICS[@]} ]}

swww init
swww img ${DIR}/${RANDOMPICS} --transition-type grow --transition-pos "$(hyprctl cursorpos)" --transition-fps 60 --transition-duration 3

flick0:
add `--invert-y` flag

inverts the y value that is received from the transition-pos flag

fixes #127

### DIFF
--- a/daemon/src/processor/animations.rs
+++ b/daemon/src/processor/animations.rs
@@ -41,6 +41,7 @@ pub struct Transition {
     pos: Position,
     bezier: BezierCurve,
     wave: (f32, f32),
+    invert_y: bool,
 }
 
 /// All transitions return whether or not they completed
@@ -70,6 +71,7 @@ impl Transition {
                 },
             ),
             wave: transition.wave,
+            invert_y: transition.invert_y,
         }
     }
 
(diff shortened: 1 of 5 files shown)


Horus645:
You work fast, @flick0! Can you please add an entry for this new flag in our documentations in [`doc/swww-img.1.scd`](https://github.com/Horus645/swww/blob/main/doc/swww-img.1.scd)?

PR #163, opened 2023-08-16
Fix wrong env variable name in documentation

mausworks:
[The code looks for SWWW_TRANSITION (and not SWWW_TRANSITION_TYPE)](https://github.com/Horus645/swww/blob/517fbeb0f831d43d6c88dac22380536b00e7d9f1/src/cli.rs#L286).

Not sure if the code is wrong, or the documentation is wrong. Either way, this is the way it works now.


Horus645:
Thanks!

Issue #165, opened 2023-08-28
Cant compile swww-deamon


Fish4203:
```
warning: generated shell completion file: "completions/swww.bash"
warning: generated shell completion file: "completions/_swww"
warning: generated shell completion file: "completions/swww.fish"
warning: generated shell completion file: "completions/swww.elv"
   Compiling rand v0.8.5
   Compiling fast_image_resize v2.7.3
   Compiling sd-notify v0.4.1
   Compiling utils v0.8.1 (/home/fish/Downloads/swww-0.8.1/utils)
   Compiling swww-daemon v0.8.1 (/home/fish/Downloads/swww-0.8.1/daemon)
error[E0658]: use of unstable library feature 'once_cell'
  --> daemon/src/main.rs:25:21
   |
25 |         Arc, Mutex, OnceLock,
   |                     ^^^^^^^^
   |
   = note: see issue #74465 <https://github.com/rust-lang/rust/issues/74465> for more information

error[E0658]: use of unstable library feature 'once_cell'
  --> daemon/src/main.rs:65:20
   |
65 | static POLL_WAKER: OnceLock<RawFd> = OnceLock::new();
   |                    ^^^^^^^^^^^^^^^
   |
   = note: see issue #74465 <https://github.com/rust-lang/rust/issues/74465> for more information

error[E0658]: use of unstable library feature 'once_cell'
  --> daemon/src/main.rs:65:38
   |
65 | static POLL_WAKER: OnceLock<RawFd> = OnceLock::new();
   |                                      ^^^^^^^^
   |
   = note: see issue #74465 <https://github.com/rust-lang/rust/issues/74465> for more information

error[E0658]: use of unstable library feature 'once_cell'
  --> daemon/src/main.rs:65:38
   |
65 | static POLL_WAKER: OnceLock<RawFd> = OnceLock::new();
   |                                      ^^^^^^^^^^^^^
   |
   = note: see issue #74465 <https://github.com/rust-lang/rust/issues/74465> for more information

error[E0658]: use of unstable library feature 'once_cell'
  --> daemon/src/main.rs:68:61
   |
68 |     if let Err(e) = nix::unistd::write(*unsafe { POLL_WAKER.get().unwrap_unchecked() }, &[0]) {
   |                                                             ^^^
   |
   = note: see issue #74465 <https://github.com/rust-lang/rust/issues/74465> for more information

error[E0658]: use of unstable library feature 'once_cell'
   --> daemon/src/main.rs:134:52
    |
[...]

PR #167, opened 2023-08-28
Add msrv to Cargo.toml

Akida31:
The `rust-version` field ensures that older rust compilers emit an error with a clearer message that a newer compiler is required.

See https://doc.rust-lang.org/cargo/reference/manifest.html#the-rust-version-field

Error message:
```
error: package `swww v0.8.1 (/path/to/swww)` cannot be built because it requires rustc 1.70 or newer, while the currently active rustc version is 1.68.2
````

see #165


Horus645:
Oooh! I didn't know that, thanks!

PR #177, opened 2023-10-08
docs: fix TRANSITION_TYPE in README

Diaoul:
The actual name of the variable is SWWW_TRANSITION and not SWWW_TRANSITION_TYPE for transition type.


Diaoul:
Code [here](https://github.com/Horus645/swww/blob/main/src/cli.rs#L286)

Horus645:
thanks!

PR #178, opened 2023-10-20
Example script for random wallpaper easy configuration

Natzgun:
For some reason the example script did not work correctly for me, plus I wasted time looking for a simple script, so I made my own, if you like you can review it and place it in the scripts section if not it is ok.


Horus645:
What didn't work in the current example scripts? Can you give me more details?

Natzgun:
It may sound silly on my part but I couldn't load the wallpaper path correctly, I couldn't find any errors in your script, perhaps due to my lack of experience that's why I made an alternative but not a replacement. :')

PR #187, opened 2023-12-16
image_pad: fix capacity of Vec

MichaelOultram:
This function will try to fill the Vec padded up to padded_h * padded_w * 3, but the initial capacity was set to padded_w * padded_w * 3. This looks to be a small typo.


Horus645:
Thanks!

Issue #111, opened 2023-04-06
FEATURE_REQUEST add support for animated webp


beh-10257:
since webp format accepts animation why not use it + the file size is considerably smaller

Horus645:
This would need to be supported by the `image` crate for us to use it, but it should be fairly easy.

beh-10257:
@Horus645 there are a lot of talk about the animated webp in the image crate is it already supported
exemple of issues
https://github.com/image-rs/image/issues/1893
+
there is this pull request I guess it's already supported

AdsonCicilioti:
And WebP static? Looks a good start.

Horus645:
@AdsonCicilioti it seems the image crate has gained support for WebP static, so it should on next release when I bump our dependencies.

PR #189, opened 2023-12-20
Animated WebP Support + Memory Optimisations

MichaelOultram:
* Create ImgBuf enum to hold File/Stdin reader
* When animated, only decode the first frame for the make_img_request
* Change compress_frames to take frames iter instead of a GifDecoder

### DIFF
--- a/src/imgproc.rs
+++ b/src/imgproc.rs
@@ -1,7 +1,11 @@
 use fast_image_resize::{FilterType, PixelType, Resizer};
-use image::{codecs::gif::GifDecoder, AnimationDecoder, DynamicImage, RgbImage};
+use image::{
+    codecs::{gif::GifDecoder, webp::WebPDecoder},
+    AnimationDecoder, DynamicImage, Frames, ImageFormat, RgbImage,
+};
 use std::{
     fs::File,
+    io::Stdin,
     io::{stdin, BufReader, Read},
     num::NonZeroU32,
     path::Path,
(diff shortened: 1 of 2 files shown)


Horus645:
Oooh! This should close #111.

Thanks!

Overall, this LGTM, let me just test it for little bit before merging.

Horus645:
Yep, this is really great, you even cleaned up some of the code.

Thanks again!

beh-10257:
is decoding the first frame only still necessary ??
from what I can see here
https://github.com/image-rs/image/issues/1550
its already added so why not enable decoding the whole thing

MichaelOultram:
> When animated, only decode the first frame for the make_img_request

@beh-10257: This text might be slightly misleading; it was an implementation detail referring to a memory optimisation not a limitation.

There are two independent IPC requests for GIF/WebP. `make_img_request` only needs the first frame (previously it was decoding every frame) and `make_animation_request` needs every frame. As these requests are independent, they were unnecessarily decoding every frame twice which consumed a lot of memory.

TLDR: Animated WebP images should show the animation correctly at the time of this commit. I've not yet tried the latest release to check if it's still working.

beh-10257:
@MichaelOultram can you try this in your system
[output.txt](https://github.com/LGFae/swww/files/14876227/output.txt)
rename it to .webp I am just getting around github limitations here
for me its not working on latest release 0.9.1

LGFae:
We need to decode just the first frame to play out the transition from the previous image to the current one. The animation then starts after that. Changing that would be pretty involved.

This is also not the place to file for bug reports. If you are having problems with `webp` images, you should open an issue.

beh-10257:
I made an issue https://github.com/LGFae/swww/issues/260
sorry for the annoyance once again
and FWIW I really though the animation not showing up was by design that why I came here sorry once again

Issue #201, opened 2024-01-31
[Bug] Static WebP crash after v0.8.2


sakarie9:
In v0.8.2, attempting to use a static WebP image results in a crash. This issue was not present in v0.8.1.

After some test I believe the problem was introduced by 2ae9eb6, specifically the line of code at

https://github.com/LGFae/swww/blob/a4c5bdbf08f6ff1839aa76f162f540b822cabca3/src/imgproc.rs#L53-L56

which mistakenly treats all WebP images as animated WebP images, then cause the panic at src/imgproc.rs:115

Reproduce
```
swww img '/tmpfs/1.webp'
thread '<unnamed>' panicked at src/imgproc.rs:115:31:
called `Option::unwrap()` on a `None` value
```

WebP info
```
File: 1.webp
RIFF HEADER:
  File size: 392764
Chunk VP8  at offset     12, length 392752
  Width: 3840
  Height: 2160
  Alpha: 0
  Animation: 0
  Format: Lossy (1)
No error detected.
```

binarypie:
I can confirm I also have this same issue on OpenSUSE Tumblelweed swww version 0.8.2

LGFae:
Hello, I have two questions, if you guys could confirm them:

Does this also happen for gifs that are a single image? (for me, it does).

Is the image set despite the crash? (for me, it is).

PR #204, opened 2024-02-09
don't crash for not animated webp images

Akida31:
This patch checks that if the webp image is animated and if not, it will create a single static image.
Note that currently the image file will be read twice. This is unfortunate but I wasn't able to find a better way to do it.

Fixes: #201

### DIFF
--- a/src/imgproc.rs
+++ b/src/imgproc.rs
@@ -21,6 +21,13 @@ use crate::cli::ResizeStrategy;
 
 use super::cli;
 
+pub enum MaybeFrames<'a> {
+    /// Image has animated Frames (is webp or gif)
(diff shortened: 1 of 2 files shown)


LGFae:
We need to determine if this also happens with Gifs. Also, I remember writing some extra logic somewhere to deal with the case where gifs have only a single frame, which we should be able to re-use it for this.

I will investigate further between today and tomorrow, but overall I *believe* this solution might be unnecessary.

Akida31:
sounds good :+1: 
I thought that gifs are always animated, but if you want I can add a check that the gif has more than one frame.
But if you have a better solution, go ahead :)

LGFae:
We merged #205 instead.

PR #205, opened 2024-02-10
fix single image webp and gif

LGFae:
@Akida31 in the end, I've used a similar solution to yours. I also end up creating a Reader twice, though the second time I am using a simple buffered file reader, rather than the Image io.

Overall, I think I think my implementation looks a little simpler. I will probably be merging this one.

### DIFF
--- a/daemon/src/wallpaper.rs
+++ b/daemon/src/wallpaper.rs
@@ -120,7 +120,7 @@ impl Wallpaper {
                 id: AtomicUsize::new(0),
                 transition_finished: Arc::new(AtomicBool::new(false)),
             },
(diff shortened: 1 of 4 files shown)


LGFae:
Fixes: #201

Akida31:
looks good :)
I saw that it won't be possible to render animated images which are passed from stdin. I don't know if it was possible before and if somebody used this but I think (but didn't try) that it should be rather easy to make the `is_animated` check work for stdin.

LGFae:
We didn't animate stdin before so I think this should be fine.

It's not *that* easy because stdin isn't seekable (like a file).

PR #206, opened 2024-02-13
fix busyloop while waiting for WlBuffers to be released

### LINKED ISSUE #202: niri: swww gets stuck in infinite loop with 100% cpu usage
Looks like `swww` does not work with the double buffering implementation in https://github.com/YaLTeR/niri. It manages to draw a single frame on startup before it gets stuck in an infinite loop consuming 100% cpu.

LGFae:
We changed from SlotPool to Multipool, which let us fix the busyloop. Suggested by  @YaLTeR.

Fixes #202.

### DIFF
--- a/daemon/src/animations/mod.rs
+++ b/daemon/src/animations/mod.rs
@@ -140,14 +140,17 @@ impl Animator {
                             continue;
                         }
 
-                        if !wallpapers[i].canvas_change(|canvas| frame.unpack(canvas)) {
-                            error!("failed to unpack frame, canvas has the wrong size");
+                        let (success, buffer) =
+                            wallpapers[i].canvas_change(|canvas| frame.unpack(canvas));
+
+                        if !success {
+                            error!("failed to unpack frame, canvas is smaller than expected");
                             wallpapers.swap_remove(i);
                             tokens.swap_remove(i);
                             continue;
                         }
 
-                        wallpapers[i].draw();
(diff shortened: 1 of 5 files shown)


LGFae:
Just finished testing with [niri](https://github.com/YaLTeR/niri). It appears to be working.

YaLTeR:
Just gave it a try, while it seems to work, the damage tracking or something appears to be broken? There are some buffer glitches.

YaLTeR:
https://github.com/LGFae/swww/assets/1794388/61c1efea-9b09-4549-9176-71d3bdb77b96

LGFae:
Jesus Chirst. I wonder why that didn't show up over here? I tested it inside River, maybe I should have spawned a new tty. I'll do that now.

LGFae:
Hmm. I am actually incapable of reproducing that. I thought this would damage the entire surface:

```rust
        let surface = self.layer_surface.wl_surface();
        surface.attach(Some(buf), 0, 0);
        surface.damage_buffer(0, 0, width, height);
        surface.commit();
```

So, because we damage the entire surface every time, I don't think it's the damage tracking...

YaLTeR:
Yeah, weird.. Maybe something to do with buffer format?

LGFae:
The buffer format didn't change. It remains Xrgb8888, which is guaranteed to exist by the protocol.

PR #209, opened 2024-02-16
New compression functions

LGFae:
This overhauls the current api for compressing animated frames.

The new api is more explicit about errors, does better checking for preconditions, and is faster.

We've dropped `lzzzz` in favor of just linking the lz4 system library directly. lz4 was already listed as one of our dependencies, so this should be fine.

We've also included hand-written SIMD implementations to further speed up both compression and decompression. Note that LZ4 takes the most amount of time when it comes to compression and decompression, so, even though this is nice, it won't make a world of difference.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -643,12 +643,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "lazy_static"
-version = "1.4.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e2abad23fbc42b3700f2f279844dc832adb2b2eb069b2df918f455c4e18cc646"
-
-[[package]]
 name = "lebe"
 version = "0.5.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1414,7 +1408,6 @@ name = "utils"
 version = "0.8.2-master"
 dependencies = [
  "criterion",
- "lazy_static",
  "lzzzz",
  "rand",
  "rkyv",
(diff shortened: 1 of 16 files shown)


LGFae:
Thanks a lot for pointing that out, @Akida31.

Interestingly, this is a problem that also existed in the previous implementation! It just hadn't blown up yet.

I've gone ahead and added some `debug_assertions` to make sure this shows up in future tests. I might add even more assertions before the final merge.

Akida31:
I ran miri after that and it looks like the errors are gone now.
Sadly miri takes really long to run all tests, on my computer like 30 minutes. So I think it is not really feasible to run miri in CI right now. But it could be worth trying to reduce the number iterations for miri so that this will be possible.
Some other thoughts to this PR:
* If somebody has the energy and time I think it would be really good to review all `unsafe` code and write [safety](https://std-dev-guide.rust-lang.org/policy/safety-comments.html) [comments](https://rust-lang.github.io/rust-clippy/master/index.html#/undocumented_unsafe_blocks) and document the assertions (e.g. `utils/src/compression/mod.rs:77`) which are necessary for soundness and may not be removed.
* I think there are some functions unnecessarily public which can be private (for example in `utils/src/compression/comp/mod.rs`)
* I think the simd checking macro/ functions could use `OnceCell` and not a `static mut` removing some unsafe code.

But these are just my thoughts and not blocking concerns for this PR, feel free to decide which of these points are good and which to ignore.
All in all I think the speedup will be nice :)

LGFae:
I think your 1st and 2nd point are very valid, and I've gone ahead and document most of the usage of unsafe and their soundness requirements.

Regarding your 3rd point, I just don't like `OnceCell` very much. It's a silly personal preference that might change in the future, but I'll leave it like this for now.

Thanks a lot for your guidance and reviews!

PR #216, opened 2024-02-21
Added Makefile and using new_unchecked instead of new on NonZero types

unixpariah:
1. **Added a Makefile**: This addition will help automate build processes, making it easier to compile and test the project.

2. **Updated NonZero types**:  The 'new' method has been replaced with 'new_unchecked' to bypass unnecessary checks. This method is unsafe as it does not verify if the value is zero, but in this context, none of the values should be zero.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -36,9 +36,9 @@ checksum = "4b46cbb362ab8752921c97e041f5e366ee6297bd428a31275b9fcf1e380f7299"
 
 [[package]]
 name = "anstream"
-version = "0.6.7"
+version = "0.6.12"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4cd2405b3ac1faab2990b74d728624cd9fd115651fcecc7c2d8daf01376275ba"
+checksum = "96b09b5178381e0874812a9b157f7fe84982617e48f71f4e3235482775e5b540"
 dependencies = [
  "anstyle",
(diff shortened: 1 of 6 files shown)


LGFae:
Sorry, but I don't like either of these changes.

The Makefile addition makes no sense; you've simply wrapped the more well know cargo commands behind an extra layer of abstraction. Furthermore, you've hard-coded the installation paths, with `su`.  This is not a good approach for package maintainers. There may be distros who want to copy these files elsewhere, and probably without using `su` directly like this.

I only use unsafe for either truly performance sensitive code (like the compression/decompression functions, because they are the main drive of high cpu usage during animation), or because I simply can't do it without unsafe (like in some instances when using the `rkyv` library. Using unsafe in these NonZero types in these instances isn't useful, because: 
  * the compiler will most likely be able to prove most of these instances are always correct, and optimize the extra checks away
  * if ever one day these things cease to be true, the code will explode in our face, like we want. With `unsafe` we would suddenly have a really annoying to track bug
  
You see, right now, the necessary invariants trivially hold (the values aren't zero). But, upon rewriting them with `unsafe`, they must how trivially hold **forever**. Even if we go through large refactors, or if the upstream libraries change, or if some unforeseen circumstance arise, etc. For this reason, it is not a good idea to use unsafe functions even if right now you can prove they are correct; because from now on, you would have to keep proving that every time the code changes.

PR #218, opened 2024-02-21
fix: process swww args even with --no-daemon

musjj:
This PR makes it so that swww args gets processed even when you pass `--no-daemon` to it. This ensures that your cached wallpaper will get loaded when you run `swww init --no-daemon`.

Fix: https://github.com/LGFae/swww/issues/217

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,5 +1,10 @@
 use clap::Parser;
-use std::{os::unix::net::UnixStream, path::PathBuf, process::Stdio, time::Duration};
+use std::{
+    os::unix::net::UnixStream,
+    path::PathBuf,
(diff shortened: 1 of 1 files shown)


LGFae:
The check failing is a false positive; don't worry about it.

musjj:
Never mind, this doesn't work when run as a service on startup. Apparently, this is a known issue: https://github.com/LGFae/swww/issues/144#issuecomment-1638335866.

For now I'm closing it, because this will be a regression for those who are using it as a service.

musjj:
Yeah, for now I think a better solution would be a command to load the cache explicitly: https://github.com/LGFae/swww/issues/219

Issue #220, opened 2024-02-21
Graphical glitches throughout half of the wallpaper


Alawapr:
This happens on half of the wallpaper in my main monitor and the entire wallpaper in my secondary monitor

https://github.com/LGFae/swww/assets/93432929/058b4183-fe91-4cec-a7f8-a6ff8840b495

Relevant info: running on Hyprland under Fedora 39 and using an Nvidia card (GTX 1660 SUPER)

0bCdian:
Having this same issue, and transition glitches on an amd 5300u (vega graphics) so it's definitely not nvidia related.

LGFae:
Just to confirm, in this on 0.8.2 or the latest master?

0bCdian:
https://github.com/LGFae/swww/assets/101421807/1e8deaa0-2c8b-4db0-93b3-6e431fed9bfd

this is on latest master, the weird glitching happens consistently when trying to set two images a the same time (im calling swww img once for each monitor with different images), this didn't happen on previous versions.

Alawapr:
> Just to confirm, in this on 0.8.2 or the latest master?

as [0bCdian](https://github.com/0bCdian) said and I'm confirming: latest master

0bCdian:
@LGFae Just informing that I tested your pull request and it fixed the weird glitching. The only visual bug persisting is the jagged lines/glitching in the transition animation, I'll be looking at the changes and testing them 👍🏾

PR #221, opened 2024-02-22
Implement frame callbacks and a new memory pool

### LINKED ISSUE #208: Weird buffer visual glitches
[](https://github.com/LGFae/swww/assets/1794388/61c1efea-9b09-4549-9176-71d3bdb77b96)

_Originally posted by @YaLTeR in https://github.com/LGFae/swww/issues/206#issuecomment-1942127690_

LGFae:
Instead of using multi pools, we are double buffering our WlBuffer so that we don't have to busy loop forever.

Fixes #208 and #220.

### DIFF
--- a/daemon/src/wallpaper.rs
+++ b/daemon/src/wallpaper.rs
@@ -4,7 +4,7 @@ use std::{
     num::NonZeroI32,
     sync::{
         atomic::{AtomicBool, AtomicUsize, Ordering},
-        Arc, Mutex, MutexGuard, RwLock, RwLockReadGuard, RwLockWriteGuard,
+        Arc, Condvar, Mutex, RwLock,
     },
 };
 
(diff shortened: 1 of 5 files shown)


jeLee6gi:
This fixes #208 for me

LGFae:
Thanks, but unfortunately right now this is not really obeying the protocol properly. I am fixing that right now.

jeLee6gi:
I tested this in niri with two outputs of different size (and different refresh rates fwiw) and everything seemed to worked just fine, cheers!

0bCdian:
I tested this pr and everything seems to be working, no more glitches!

daemon/src/wallpaper.rs
```diff
@@ -31,45 +28,57 @@ struct AnimationState {
...
+    pub(super) fn set_transition_done(&self, wallpaper: &Wallpaper) {
         if wallpaper.has_animation_id(self) {
             self.transition_done.store(true, Ordering::Release);
         }
     }
 }
 
+struct FrameCallbackHandler {
+    cvar: Condvar,
+    time: Mutex<Option<u32>>,
```
YaLTeR, line 50: It's probably worth adding a comment that this `time`, that is received in the frame callback, doesn't really mean anything and in particular should not be used for any kind of frame timing. (For that purpose, you can use presentation-time + some logic to predict when a frame will be presented.)

YaLTeR:
Gave this a quick test on my setup, seems to work!

daemon/src/wallpaper.rs
LGFae: Ah, good point.

PR #223, opened 2024-02-24
add restore command

musjj:
This allows the user to manually restore the cache via a subcommand.

Played around with it for a bit, and it seems to work fine. Let me know if there's anything wrong.

Fix: https://github.com/LGFae/swww/issues/219


LGFae:
This looks very cool. Thanks!

Issue #144, opened 2023-07-14
Black background on login since 0.8.0 / 0.8.1


Yasuhan:
Hey

ever since the latest update (both 0.8 and 0.8.1) on login in to my session I am just greeted by a black background, instead of my own. After either manually using swww img or simply reloading my hyprland config the wallpaper appears though. 

This is what I got in my hyprland.conf

`exec = swww init`  

`exec = swww img ~/.config/hypr/bg.jpg`

And I don't think there's anything wrong with that, as simply reloading said config makes the bg appear as intended.

So far I've deleted the .cache/swww folder several times, as well as the swww.socket under /var/run.  Any other ideas how to fix this issue?

Horus645:
I have one idea: set an image, don't delete the cache, comment out your second `exec`, then try again. `swww` should automatically load the last image that was set before logging out.

Although, this configuration of yours should be working, specially if it worked before.

Yasuhan:
Sadly neither of those options worked for me. I'm just really confused because it worked before the update.

o-dasher:
Just bumping this to say the same is happening to me since like a week or so.

Horus645:
> Just bumping this to say the same is happening to me since like a week or so.

If it has been happening for a week, then it could be something other than `swww`, since the update were released just two days ago.

OopsYao:
Can confirm. I use `swww init --no-daemon` as a systemd service (as `swww init` does not work). And it seems that `swww init --no-daemon` (even run in a terminal without systemd) does not respect cache after v0.8.1.

I didn't try v0.8.0, and `swww init` does respect the cache if run in a terminal.

Hyprland on Arch linux with ker 6.4.3-arch1-1

hekatonkheirex:
> 
> EDIT:
> 
> Another idea is to merge those two commands into one:
> 
> ```
> exec = swww init && swww img ~/.config/hypr/bg.jpg
> ```
> 

I can confirm I have this command in my config and the wallpaper isn't loading at startup.

Also using latest Arch + Hyprland.

PlasticSoup:
Also Arch+Hyprland. Maybe the issue lies there? I have always called a script from hyprland.conf to start swww with a random picture. All images in "wallpapr" are animated gifs, I never tried loading static images or colors.

`swww init && swww img $(find ~/Pictures/wallpapr -type f | shuf -n 1)`

[...]

PR #225, opened 2024-02-24
wait for daemon initialization before certain requests

musjj:
`swww {init,img,restore,clear}` now waits for the daemon to initialize before making the request.

Not sure if this is the best approach, but it seems to work ok.

Fix: https://github.com/LGFae/swww/issues/224

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -49,20 +49,18 @@ fn main() -> Result<(), String> {
         }
     }
 
-    if let Swww::Init { .. } | Swww::Img(_) | Swww::Restore(_) | Swww::Clear(_) = &swww {
-        let mut configured = false;
-        while !configured {
-            let socket = connect_to_socket(5, 100)?;
-            Request::Init.send(&socket)?;
-            let bytes = read_socket(&socket)?;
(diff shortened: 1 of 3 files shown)


LGFae:
I wonder if this also fixes #144. It seems like it would.

src/main.rs
```diff
@@ -49,6 +49,22 @@ fn main() -> Result<(), String> {
         }
     }
 
+    if let Swww::Init { .. } | Swww::Img(_) | Swww::Restore(_) | Swww::Clear(_) = &swww {
```
LGFae, line 52: I don't think we need this if statement. We should **always** make sure the daemon is running before sending in any requests to it.

LGFae (changes requested):
I like this idea a lot. The only thing besides what I've commented above is that maybe you could rename Request::Init to Request::Ping (this will also trigger a rename in its ArchivedRequest equivalent)? This would be more accurate to what we are doing now.

If it's too much work I can do it after merging too.

src/main.rs
musjj: That sounds like a good idea!

PR #226, opened 2024-02-28
Use 3channel format

LGFae:
Using `Format::Bgr888` or `Format::Rgb888` when they are available result in memory savings (since our buffers have 3/4 the size) and slight performance improvements, because we don't have to swap the R and B channels.

This PR increases the code base's complexity somewhat. It might be worth it to spend some trying to make things clearer and/or simpler in general.

### DIFF
--- a/daemon/src/main.rs
+++ b/daemon/src/main.rs
@@ -104,6 +104,7 @@ extern "C" fn signal_handler(_s: i32) {
 fn main() -> Result<(), String> {
     rayon::ThreadPoolBuilder::default()
         .thread_name(|i| format!("rayon thread {i}"))
+        .stack_size(1 << 18) // 256KiB; we do not need a large stack
         .build_global()
         .expect("failed to configure rayon global thread pool");
     make_logger();
--- a/daemon/src/main.rs
+++ b/daemon/src/main.rs
@@ -76,22 +76,22 @@ static PIXEL_FORMAT: OnceLock<PixelFormat> = OnceLock::new();
 #[inline]
 pub fn wl_shm_format() -> wl_shm::Format {
     debug_assert!(WL_SHM_FORMAT.get().is_some());
-    // SAFETY: this is safe because we initialize it in Daemon::new, before we ever call this in
-    // the wallpaper structs
-    *unsafe { WL_SHM_FORMAT.get().unwrap_unchecked() }
+    *WL_SHM_FORMAT.get().unwrap_or(&wl_shm::Format::Xrgb8888)
 }
 
 #[inline]
 pub fn pixel_format() -> PixelFormat {
     debug_assert!(PIXEL_FORMAT.get().is_some());
-    // SAFETY: this is safe because we initialize it in Daemon::new, before we ever call this in
-    // the wallpaper structs
-    *unsafe { PIXEL_FORMAT.get().unwrap_unchecked() }
+    *PIXEL_FORMAT.get().unwrap_or(&PixelFormat::Xrgb)
 }
 
 #[inline]
 pub fn wake_poll() {
     debug_assert!(POLL_WAKER.get().is_some());
+
+    // SAFETY: POLL_WAKER is set up in setup_signals_and_pipe, which is called early in main
+    // and panics if it fails. By the time anyone calls this function, POLL_WAKER will certainly
+    // already have been initialized.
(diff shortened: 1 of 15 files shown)


utils/src/compression/decomp/mod.rs
```diff
@@ -58,3 +58,46 @@ pub(super) fn unpack_bytes(buf: &mut [u8], diff: &[u8]) {
         pix_idx += 1;
     }
 }
+
```
Akida31, line 61:
```suggestion

#[cfg(test)]
mod tests {
    #[test]
    fn test_ub() {
        let mut bytes = vec![u8::MAX; 9];
        let diff = vec![u8::MAX; 18];

        super::unpack_bytes_3channels(&mut bytes, &diff);
    }
}
```
If you try this test with miri (`cargo +nightly miri test test_ub`) it will complain.
I wasn't able to fix this easily (adding `diff_idx < len &&` to lines 73, 81 and before 85 did not help).
I am not sure if this can happen in practice (e.g. a malicious ipc request) but it may make sense to add a check or make this function unsafe, document the safety invariant and add the check in the calling function. I think this may be a problem elsewhere but saw this in the PR.
I think checking at the beginning of the function that the input has only `u8::MAX` at the correct position should not make a performance difference but I didn't try this.

LGFae: This should never happen. We only call this function in `Decompressor::decompress` and `Decompressor::decompress_archived`, which receives a `BitPack`, that can only be created by the `Compressor`. So the inner `diff` slice can never be wrong.

> a malicious ipc request

This would indeed fuck us up :sweat_smile:. They would have to match `rkyv`'s specific inner format. Which I guess would be possible if they copy pasted some code from our code base. But this is could also happen with our current implementation in the main branch, so it would have to be a more fundamental fix.

> I think checking at the beginning of the function that the input has only u8::MAX at the correct position should not make a performance difference but I didn't try this.

Mm... I wonder. Here's the thing, when decompressing, we first check how many pixels we skip. If we read a `u8::MAX`, it means we need to read the next byte and sum to the current value. After we read the first non `u8::MAX`, we add it to the value we have, and skip that many pixels. Then, we go through the same process, but now we are trying to get the number of pixels to copy over. Then we copy the bytes and restart, until there's nothing left to read from `diff`.

So, there are many valid places for `u8::MAX`, checking it might take a while.

The solution might honestly be just not using `unsafe`, but that will be unfortunate since I got a measurable performance increase writing the function this way, and that's important because this runs during animation, so it will continuously consume the CPU's resources.

As an alternative, we might be capable of adding some carefully-chosen manual checks that will prevent this from happening, while keeping most of the performance gains.

daemon/src/main.rs
```diff
@@ -64,8 +70,28 @@ fn should_daemon_exit() -> bool {
...
+    // the wallpaper structs
+    *unsafe { WL_SHM_FORMAT.get().unwrap_unchecked() }
+}
+
+#[inline]
+pub fn pixel_format() -> PixelFormat {
+    debug_assert!(PIXEL_FORMAT.get().is_some());
+    // SAFETY: this is safe because we initialize it in Daemon::new, before we ever call this in
+    // the wallpaper structs
+    *unsafe { PIXEL_FORMAT.get().unwrap_unchecked() }
```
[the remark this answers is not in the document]
LGFae: You are right. I will use your suggestion.

utils/src/compression/decomp/mod.rs
LGFae: I think  I've managed to prevent UB using just 2 assertions. The first, before we start decompressing, to make sure `diff` always ends with two `0`s. If it does, we will never read out-of-bounds while parsing the `u8::MAX`.

Second, I just assert once before copying the bytes over that `diff` has the right size.

Besides that, I've also added an extra check to make sure the `lz4` compressed `BitPack` is correct, since that could also be a problem.

Akida31: I _think_ these changes improve are good, but I can't know for sure. Having potential UB on malicious use is a bit sad but I think it is okay, if you had written swww in c it would certainly be a different situation.
There are now some asserts in (looking from far) hotter loops but if you've checked that the performance stays the same, very good!

PR #230, opened 2024-03-09
Drop daemonizing logic

WhyNotHugo:
Run in foreground, which is what service managers expect. Usages like running `exec swww` in sway's config continue to work fine. User session start-up scripts can simply run `swww &`.

Fixes: https://github.com/LGFae/swww/issues/182


src/main.rs
```diff
@@ -14,46 +14,6 @@ use cli::{ResizeStrategy, Swww};
...
-                        socket_path.to_string_lossy()
-                    );
-                    if let Err(e) = std::fs::remove_file(socket_path) {
-                        return Err(format!("failed to delete previous socket: {e}"));
-                    }
-                }
-            }
-            Ok(true) => {
-                return Err("There seems to already be another instance running...".to_string())
-            }
```
WhyNotHugo, line 39: I think I actually need to move this logic into the daemon, right?

LGFae: It's already there, actually.

doc/swww-img.1.scd
```diff
@@ -218,8 +218,7 @@ The images sent will be cached at _$XDG_CACHE_HOME/swww_ or _$HOME/.cache/swww_
 if $XDG_CACHE_HOME does not exist. For each monitor, there will be a file in
 those locations corresponding to the current image/animation being displayed.
 Importantly, **cache will only be loaded during initialization if you use swww
-init**. That is, calling `swww-daemon` directly will **NOT** load the cache, but
-calling `swww-init` will.
+init**. That is, calling `swww-daemon` directly will **NOT** load the cache.
```
LGFae, line 221: Ah, this isn't true anymore. `swww-daemon` loads the cache now, unless you pass it the `--no-cache`  flag.

src/main.rs
LGFae: It's already there, actually :).

LGFae:
From what I can tell, this looks done. Are you planning on doing anything else?

Issue #233, opened 2024-03-16
Wallpaper show distorted and in grayscale


Hanazono-Yurine:
Buenos Dias meus consagrados.

Today I performed a complete update on my system. Then I restarted my laptop.
But when the system start, my wallpaper was distorted and in grayscale.
I tried changing to another wallpaper, but it stayed the same, so I tried set to a random color with `swww clear a260f1`, but it stayed in grayscale.
I installed the `swww-git` package with yay.

Some info:

Screenshot
![screenshot-2024-03-16_17:09:42](https://github.com/LGFae/swww/assets/142830863/8405a310-8601-4d8f-a13c-349b11d11f57)

```
$ swww --version
swww 0.8.2-masterV2
```
The version on `pacman -Qi swww-git`, is the last commit of the main branch
```
0.4.0.r301.g0516ea0-1
```

```
$ swww query
eDP-1: 1366x768, scale: 1, currently displaying: image: /home/zoe/Pictures/Wallpapers_Desktop/Games/381583.jpg
```

I'm using Arch with the latest kernel version `6.7.9-arch1-1`
If you need more information, just let me know and I'll do my best to provide it.

LGFae:
Can you try running `swww-daemon` manually in a different terminal, to see what `wl_shm format` it is using? I feel this might be a problem with that.

Hanazono-Yurine:
I'm using hyprland, more specifically `hyprland-git`.
Version of hyprland if necessary:
```
$ hyprctl version
Hyprland, built from branch main at commit 0e87a08e15c023325b64920d9e1159f38a090695 dirty (renderer: disable surface adjustments for misaligned reported when manual resizing).
Date: Sat Mar 16 14:56:09 2024
Tag: v0.37.1-4-g0e87a08e

flags: (if any)
```
The output of `swww-daemon` :
```
$ swww-daemon
22:12:30 [INFO] (main) Initialization succeeded! Starting main loop...
22:12:30 [INFO] (main) Selected wl_shm format: Bgr888
```
And about my monitor setup, maybe I'm dumb, but I don't know if that's what you want, so I just copied my setup from hyprland
`monitor=,highres,auto,1`
I'm on my laptop, without any external monitor if that's what you wanted.

LGFae:
Hm. I have been unable to reproduce this with the `hyprland-git`. Try the following:

```
swww kill # kill any currently running instance
swww clear-cache # delete the cache; it might be causing problems
swww init # init it again
[...]

PR #235, opened 2024-03-22
option to force use of a wayland_shm format

LGFae:
This should aid in debugging and serve as a workaround if there's something wrong with the non-default formats.

This is intended as a workaround for #233.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,6 +10,14 @@ edition = "2021"
 # we use `std::cell::OnceLock` which was stabilized in 1.70
 rust-version = "1.70"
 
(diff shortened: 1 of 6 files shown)


skinnypuppy:
Just tested on my laptop, and running swww-daemon --format xrgb now works again. Just goofing around and tried swww init --format xrgb to see if that would work without tieing up a terminal and that was a no-go, right back to greyscale skewed image.

LGFae:
> tried swww init --format xrgb

Yeah, my bad, I forgot make the client forward the option to the daemon :sweat_smile:

Issue #182, opened 2023-11-06
Don't run in background by default


WhyNotHugo:
By default, `swww` forks and runs as a background process. Every time somebody runs `swww init`, the output of the command is not visible, and killing or restarting the command isn't as simple as <kbd>Ctrl</kbd>+<kbd>c</kbd>. One needs to inspect running processes and figure out which is the right one. There is no obvious indicator of what's going on, and this is especially worse on the first usage of `swww`, where it seems like it's exiting immediately and not really working properly.

This runs contrary to what the gran majority of other commands do: simply run in foreground. If somebody needs to run it in background, popular shells (including `bash`, `zsh`,  and `ksh`) allow simply using `swww init &` to run in background.

Service managers and supervisors also always expect services to run in foreground; demonising like this makes the job of service supervisors a lot harder, since tracking the lifetime of the child doesn't track the lifetime of the service. Eventually the solution is to always  use `swww init --no-daemon`.

I'd like to propose simply running in foreground like any other command and dropping the extra code for daemonisation. For any unique situation where running in background is preferible, using `swww init &` (or maybe `nohup swww init &`) is a trivial workaround.

Horus645:
Interesting proposal. It seems to me like a big breaking change, considering the way we've done things since the beginning. I'll think about it, but I feel like this is a good idea overall.

Nemo157:
A semi-related question: is `swww-daemon` public API? For my systemd service I run that directly instead of going through `swww init --no-daemon` to avoid the extra layering.

Horus645:
Running the daemon directly is currently exactly equivalent to running `init --no-daemon`. I cannot guarantee this will be the case forever, though.

Nevertheless, it would be a breaking change, should it ever change, with big bold letters in the release page.

WhyNotHugo:
Having `swww init` an `swww-daemon` do the same thing is kind of ambiguous.
What do you think of merging both commands into one? Or maybe drop `swww init`?

KAGEYAM4:
> Having `swww init` an `swww-daemon` do the same thing is kind of ambiguous. What do you think of merging both commands into one? Or maybe drop `swww init`?

`man swww-daemon` says this is for debug pupose, and `swww init` should be preffered as it waits for the daemon to be ready.

But i agree the, name and functionality is ambigous unless one looks at the man-page.

KAGEYAM4:
[...]

PR #238, opened 2024-03-23
Deprecate swww init

LGFae:
This is the first step towards #182. For the next `swww` version, we deprecate the `init` command. In the version following that we will remove it completely.


LGFae:
Gah, I commited the changes to the CHANGELOG in the wrong branch. Well, whatever.

PR #254, opened 2024-03-28
Fix build on i686 (32bit x86)

Calandracas606:
The build was failing in i686 (32bit x86) because the module `std::arch::x86_64` is only available on x86_64. i686 needs to be detected, so that the module `std::arch::x86` is used instead.


LGFae:
Thanks!

PR #273, opened 2024-04-10
Add support for running on multiple Wayland displays simultaneously.

### LINKED ISSUE #272: Allow running in nested Wayland sessions
If you run a nested Wayland WM, then try running ``swww-daemon`` inside that one, it'll try and rebind to the socket that an outer instance is using. 

I think the easiest way to implement this would be to make the socket filename `swww-$WAYLAND_DISPLAY.socket`. 

I understand if this isn't in-scope for the project, but if it is, I'm willing to implement it myself.

Fuyukai:
This changes the socket that ``swww`` listens to to ``/run/user/$USERID/swww-$WAYLAND_DISPLAY.socket``, meaning that if two instances are running simultaneously, they don't conflict.

I also adjusted the daemon checking logic so that it scans for the file descriptor of the socket file open, so that two daemons won't get confused about each other.

This is my first substantial Rust PR; apologies if there's code issues or non-idiomatic things. I tried to keep to the code style of the project as much as possible.

Closes #272.

### DIFF
--- a/daemon/src/main.rs
+++ b/daemon/src/main.rs
@@ -17,7 +17,6 @@ use wallpaper::Wallpaper;
 
 use std::{
     fs,
-    io::{BufRead, BufReader},
     os::{
         fd::OwnedFd,
         unix::net::{UnixListener, UnixStream},
@@ -44,7 +43,9 @@ use wayland_client::{
     Connection, Dispatch, Proxy, QueueHandle,
 };
 
-use utils::ipc::{get_socket_path, Answer, BgInfo, PixelFormat, Request};
+use utils::ipc::{
+    connect_to_socket, get_socket_path, read_socket, Answer, BgInfo, PixelFormat, Request,
+};
 
 use animations::Animator;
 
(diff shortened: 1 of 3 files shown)


daemon/src/main.rs
```diff
@@ -725,7 +734,81 @@ fn make_logger(quiet: bool) {
...
+
+    Ok(false)
+}
+
+fn is_daemon_running(addr: &PathBuf) -> Result<bool, String> {
+    let fd_num = match find_fd_number(addr)? {
+        FdFindResult::GotIt(n) => n,
+        FdFindResult::NobodysListening => return Ok(false),
+    };
+
```
LGFae, line 811: I see you had to solve a different problem to get this work (the problem of detecting whether a daemon is already running). There's an easier way to do this. If the socket path already exists, simply try Pinging the daemon (like we do [here](https://github.com/LGFae/swww/blob/main/src/main.rs#L64)). If it returns, the daemon is already running. If it doesn't, we can continue.

daemon/src/main.rs
```diff
@@ -739,12 +822,13 @@ fn is_daemon_running() -> Result<bool, String> {
...
             let mut entry_path = entry.path();
             entry_path.push("cmdline");
             if let Ok(cmd) = std::fs::read_to_string(entry_path) {
                 let mut args = cmd.split(&[' ', '\0']);
                 if let Some(arg0) = args.next() {
-                    if arg0.ends_with("swww-daemon") {
+                    if arg0.ends_with("swww-daemon") && check_fds(fd_num, entry.path())? {
                         return Ok(true);
                     }
                 }
```
LGFae, line 834: If you implement what I've commented above, this proc traversal code will become redundant, and you can delete it.

LGFae (commented):
This is pretty good, but I think there's a simpler way of doing some stuff. I actually thought this would be easier...

In any case, this is looking pretty good, especially for a first PR!

Fuyukai:
I had to move the ``connect_to_socket`` method into the shared code between the client and server; it also now explicitly takes the socket path so the daemon doesn't try checking the environment twice.

I kept the commit with the ``/proc`` scanning theree but I can rebase it out if you want.

utils/src/ipc.rs
```diff
@@ -360,3 +374,40 @@ pub fn get_cache_path() -> Result<PathBuf, String> {
...
+
+                return Ok(socket);
+            }
+            Err(e) => error = Some(e),
+        }
+        std::thread::sleep(Duration::from_millis(interval));
+    }
+    let error = error.unwrap();
+    if error.kind() == std::io::ErrorKind::NotFound {
+        return Err("Socket file not found. Are you sure swww-daemon is running?".to_string());
```
LGFae, line 409: Alright, now we have a new problem. This message will be printed every time we try to initialize the daemon when there wasn't a previous socket file there. We do not want that.

So, I'd recommend changing this function to return a `std::io::Result<UnixStream>`. Then, all the errors we are currently using `format!` on can be simplified. For example:

```rust
if let Err(e) = socket.set_nonblocking(false) {
    return Err(format!("Failed to set blocking connection: {e}"));
}
```
Can become just
```rust
socket.set_nonblocking(false)?;
```
And so on.

After that, you will need to fix the error handling in every place (4) this function is called. For most of them, I think you can just do `.map(|e| format!("failed to connect to socket: {e}")?;` on the function call.

There's one specific case, in the client's main function, where you will have to check if the error is `std::io::ErrorKind::NotFound`, and return a special message if it is.

LGFae (changes requested):
Ok, almost there. There's just one final thing to care of.

utils/src/ipc.rs
Fuyukai: > This message will be printed every time we try to initialize the daemon when there wasn't a previous socket file there.

The ``is_daemon_running`` function is only called if the socket file exists. If it doesn't, that entire block gets skipped over. The error is also suppressed if the file exists but there's no daemon listening.

I can still refacttor the error handling if you wish, but that feels like it would be out of scope for this PR...?

LGFae: Ah, I see. Indeed, you are right. I was seeing it print an error because I was testing it the wrong way.

LGFae:
Thanks for your work!

PR #286, opened 2024-04-18
Add --no-cache option to swww-daemon

lucasreis1:
With the deprecation of `swww init`, `swww-daemon` was missing a key feature of allowing users to start the daemon without caching wallpapers from previous sessions. This is a step toward fixing this issue, although this still invalidates loading the cache even for new outputs (monitor re-connections).

If you're uncomfortable with blocking caching even for newer outputs and can point me to a different approach (as I'm pretty new to rust!), I'd gladly improve the PR :)

### DIFF
--- a/daemon/src/cli.rs
+++ b/daemon/src/cli.rs
@@ -3,11 +3,13 @@ use utils::ipc::PixelFormat;
 pub struct Cli {
     pub format: Option<PixelFormat>,
     pub quiet: bool,
+    pub no_cache: bool,
 }
 
 impl Cli {
     pub fn new() -> Self {
         let mut quiet = false;
+        let mut no_cache = false;
         let mut format = None;
         let mut args = std::env::args();
         args.next(); // skip the first argument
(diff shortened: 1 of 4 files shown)


daemon/src/cli.rs
```diff
@@ -40,6 +43,12 @@ impl Cli {
                     println!("          Whatever you chose, make sure you compositor actually supports it!");
                     println!("          'xrgb' is the most compatible one.");
                     println!();
+                    println!("  --no-cache");
+                    println!(
+                        "          Don't search the cache for the last wallpaper for each output"
+                    );
+                    println!("          Useful if you always want to select which iamge `swww` loads manually using `swww img`");
```
LGFae, line 50: Typo 'iamge' -> 'image'

LGFae (changes requested):
This looks pretty good. I think it's fine to not load on monitor's reconnect for now.

In any case, there's a typo, and could you please also update the documentation in `doc/swww-daemon.1.scd` with the new option?

lucasreis1:
Done!

LGFae:
Thanks!

PR #292, opened 2024-04-24
fix switching width and height

AhJi26:
The previous fix called set_vertical before set_dimensions, causing the width and height to switch before getting their values. This PR add lock.is_vertical and swaps the weight and height in set_dimensions (after w and h receive their values) to resolve the issue. 

This PR has been tested on my machine with a vertical monitor, and it functions correctly.


clsty:
Great job, I can confirm that it works on my PC (multiple monitors, one of them vertical) too.

LGFae:
Thanks a lot!

PR #308, opened 2024-05-13
replaced deprecated swww init commands with swww-daemon

JaKooLit:
This is just to replace the deprecated swww init with swww-daemon


LGFae:
Oh shit, I thought I had done that. Thanks!

PR #313, opened 2024-05-19
Add kitty as an alternative

kovidgoyal:
kitty, with its panel kitten can be used as the desktop wallpaper with layer-shell supporting compositors. This can be used to display the output of arbitrary TUI programs/scripts as your desktop wallpaper.


LGFae:
Ooh, very interesting. Thanks!

PR #319, opened 2024-05-28
workflows: add nix-tests workflow

JohnRTitor:
This PR adds one workflow:

1. nix-tests - checks nix files, builds the package, on every push and PR
2. Updates `update-flake-lock` to run monthly, as there is no need to update dependencies weekly

Workflow-demo: [nix-tests](https://github.com/LGFae/swww/actions/runs/9265935709/job/25489044406?pr=319)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,3 +14,5 @@ jobs:
     uses: ./.github/workflows/dependencies.yml
   test:
     uses: ./.github/workflows/test.yml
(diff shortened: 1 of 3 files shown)


LGFae:
Ah, thanks. I was thinking of changing `update-flack-lock` to run monthly myself.

LGFae:
Wait, what does `nix-tests` actually do that's different from our current tests?

I should have asked this before merging, but oh well...

PR #330, opened 2024-06-18
Improve cargo workspace

rkuklik:
I moved swww client to separate directory and renamed utils to more intuitive common


Cargo.toml
```diff
@@ -1,14 +1,30 @@
 [workspace]
-members = [".", "daemon", "utils"]
-default-members = [".", "daemon"]
+resolver = "2"
```
LGFae, line 2: This is unnecessary, since the 2021 edition already uses resolver 2.

version.sh
```diff
@@ -21,16 +21,16 @@ cargo test --workspace -- --include-ignored
...
-mv -v TMP utils/Cargo.toml
+sed "s/^version = .*/version = \"$1\"/" common/Cargo.toml > TMP
+mv -v TMP common/Cargo.toml
 
 sed \
 	-e "s/^version = .*/version = \"$1\"/" \
-	-e "s/^utils.*/utils = { version = \"$1\", path = \"..\/utils\" }/" \
+	-e "s/^common.*/common = { version = \"$1\", path = \"..\/common\" }/" \
 	daemon/Cargo.toml > TMP
 mv -v TMP daemon/Cargo.toml
```
LGFae, line 36: These were updating the versions in every crate. Since now there is only one global one, you just need

```bash
sed  -e "s/^version = .*/version = \"$1\"/" Cargo.toml > TMP
mv -v TMP Cargo.toml
```

LGFae (commented):
Just two small things.

LGFae:
Very nice.

PR #331, opened 2024-06-18
Typesafe IPC

rkuklik:
This is first (and probably not last) PR for IPC refactor. For now this is a draft because I am unsure if PRs of this size are either to big or too inconsequential.

Apart from sending and receiving itself, I will take a look at some semi-duplicated code (looking at you, `RequestSend` and `RequestRecv`), though this will not be included in this PR.

Is there any form of compatibility guarantee between swww and swww-daemon if version bump ocuurs?

### DIFF
--- a/common/src/ipc/socket.rs
+++ b/common/src/ipc/socket.rs
@@ -6,17 +6,10 @@ use std::time::Duration;
 use rustix::fd::OwnedFd;
 use rustix::io::Errno;
 use rustix::net;
-use rustix::net::RecvFlags;
 
 use super::ErrnoExt;
 use super::IpcError;
 use super::IpcErrorKind;
-use super::Mmap;
-
-pub struct SocketMsg {
-    pub(super) code: u8,
-    pub(super) shm: Option<Mmap>,
-}
 
 /// Represents client in IPC communication, via typestate pattern in [`IpcSocket`]
 pub struct Client;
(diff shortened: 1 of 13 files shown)


LGFae:
No. You may break as many things as you want :)

rkuklik:
I noticed couple `cfg(target_os = "linux")` in `mmap` module. Is there a reason for this? As far as I know, wayland is official only for linux. Does this project support (or plans to support) BSDs?

rkuklik:
This should conclude first part if IPC refactor. In next PR I will reorganize daemon and client a bit to facilitate easier adjustments.

LGFae:
> I noticed couple `cfg(target_os = "linux")` in `mmap` module. Is there a reason for this? As far as I know, wayland is official only for linux. Does this project support (or plans to support) BSDs?

Wayland is already being used by some people in FreeBSD. They've even opened issues in this project.

Furthermore, in general, Wayland isn't Linux-specific. There was a time where `libwayland`  made Linux-specific syscalls in its implementation, but it's no longer the case.

common/src/ipc/socket.rs
```diff
@@ -1,141 +1,133 @@
...
+use std::sync::OnceLock;
+use std::time::Duration;
+
+use rustix::fd::OwnedFd;
+use rustix::io::Errno;
+use rustix::net;
+
+use super::ErrnoExt;
+use super::IpcError;
+use super::IpcErrorKind;
```
LGFae, line 12: Is there a reason why you are doing imports like this? I don't particularly mind, but it isn't consistent with the rest of the codebase. I supposed we could normalize them later...

LGFae (commented):
Sorry it took me so long. This is a large PR and I am kind of busy right now.


In any case, I am really liking how this is turning out. I just have a tiny nitpick regarding how you do imports.

common/src/ipc/socket.rs
rkuklik: I simply prefer this style. When I change imports, the resulting diff is on one line (with entire path). Also, I think it is easier to parse (entire path in one spot, I don't have to jump around to see who declared something like `fs` module) and edit in general (`dd` in vim).

rkuklik:
> Wayland is already being used by some people in FreeBSD. They've even opened issues in this project.

Alright. And as for `MmapedStr` and `MmapedBytes`, is there a reason for their existence? I tried to look up their usage and it seems to me that they are only really used for IPC in daemon (to avoid lifetimes?). I am not that familiar with shared memory, but I would like to get rid of them in favor of regular references when I rework serde. Sounds OK?

LGFae:
> And as for MmapedStr and MmapedBytes, is there a reason for their existence? I tried to look up their usage and it seems to me that they are only really used for IPC in daemon (to avoid lifetimes?). I am not that familiar with shared memory, but I would like to get rid of them in favor of regular references when I rework serde. Sounds OK?

They are there to avoid copies. By memory mapping them we have only one shared memory buffer between the client and the daemon. Also, we won't be using serde, at all. Serde does a bunch of unnecessary copies around, and increases our dependencies significantly due to everything we have to pull. It's really good for large applications where writing everything manually would be a pain, or simply when you don't want to bother. But I went to great lengths to make IPC as efficient as I could. Our serialization and de-serialization will remain all manual.

I believe the easiest way of making the code more readable would be by using something like a builder pattern. You wrap `RawMsg` with a `ReqBuilder`. And then you can call `builder.next_mmaped_str()`, which will do all the "ugly" stuff for you.

rkuklik:
Sorry, that's not really what I meant. By serde I meant reading and writing messages to the shared buffer (without a protocol). But when receiving and deserializing data, I noticed tho two aforementioned types being built. Is there a reason not to create `&[u8]` and `&str` directly and shove a lifetime parameter to relevant data structures? I didn't want to introduce a new dependency.

LGFae:
Ah, I get it now, my bad.

I don't 100% remember what the problem was, but I believe it was that, when we send in an animation, we will have the image bytes and the animation bytes all mapped, and I wanted to unmap just the image bytes and keep the animation bytes.

In any case, if you think you can simplify the logic by removing them, go ahead and give it a try!

PR #334, opened 2024-06-26
Fix typo in command docs

Not-Glunk:
Accidentally found it when messing around with arguments after finally getting swww to work on my machine, awesome piece of software!


LGFae:
Thanks!

PR #335, opened 2024-06-26
Finalize IPC behaviour

rkuklik:
This is probably largest part of my IPC journey. Currently very much work in progress. It will probably consist of three parts:

1. Sketch out new types and API in general and implement alongside old system.
2. Fuzz the hell out of it. Memory allocation and type safety very much depend on correctness of `Serialize` impl and is crucial to get right.
3. If two previous steps work, out with the old, in with the new.
4. Profit

Given that this will basically a rewrite of all communication, I would like opinion of other people before merging, as this is quite broad.

### DIFF
--- a/common/src/ipc/serde.rs
+++ b/common/src/ipc/serde.rs
@@ -62,17 +62,17 @@ impl<'a> Cursor<&'a [u8]> {
     }
 }
 
-trait Tagged<'a> {
+trait Tagged {
     fn tag(&self) -> u32;
 }
 
-impl<T> Tagged<'_> for [T] {
+impl<T> Tagged for [T] {
     fn tag(&self) -> u32 {
         self.len() as u32
     }
 }
 
-impl Tagged<'_> for str {
+impl Tagged for str {
     fn tag(&self) -> u32 {
         self.as_bytes().tag()
     }
@@ -175,6 +175,14 @@ impl<T: Serialize> Serialize for Vec2<T> {
     }
 }
 
+impl<'a, T: Deserialize<'a>> Deserialize<'a> for Vec2<T> {
+    fn deserialize(buffer: &mut Cursor<&'a [u8]>) -> Self {
+        let x = T::deserialize(buffer);
+        let y = T::deserialize(buffer);
+        Self { x, y }
+    }
+}
+
 impl<T: Serialize> Serialize for Box<[T]> {
     fn serialize(&self, buffer: &mut Cursor<&mut [u8]>) {
         self.tag().serialize(buffer);
@@ -193,14 +201,6 @@ impl<'a, T: Deserialize<'a>> Deserialize<'a> for Box<[T]> {
         (0..u32::deserialize(buffer))
             .map(|_| T::deserialize(buffer))
             .collect()
-    }
-}
-
-impl<'a, T: Deserialize<'a>> Deserialize<'a> for Vec2<T> {
-    fn deserialize(buffer: &mut Cursor<&'a [u8]>) -> Self {
-        let x = T::deserialize(buffer);
-        let y = T::deserialize(buffer);
-        Self { x, y }
     }
 }
 
--- a/common/src/ipc/serde.rs
+++ b/common/src/ipc/serde.rs
@@ -1,0 +1,631 @@
+use std::mem::size_of;
+use std::mem::size_of_val;
+use std::num::NonZeroI32;
+use std::num::NonZeroU8;
+use std::time::Duration;
+
+use super::types2::Animation;
+use super::types2::BitPack;
+use super::types2::ClearRequest;
+use super::types2::Coord;
+use super::types2::Image;
+use super::types2::ImageDescription;
+use super::types2::ImageRequest;
+use super::types2::Info;
+use super::types2::PixelFormat;
+use super::types2::Scale;
+use super::types2::Transition;
+use super::types2::TransitionType;
+use super::types2::Vec2;
+
+/// Type for managing access in [`Serialize`] and [`Deserialize`]
+pub struct Cursor<T> {
+    buf: T,
+    pos: usize,
+}
+
+impl<T> Cursor<T> {
+    /// Create [`Self`] from provided buffer
+    pub fn new(buf: T) -> Self {
+        Self { buf, pos: 0 }
+    }
+
+    /// Extract buffer back, discarding all internal state.
+    pub fn finish(self) -> T {
+        self.buf
+    }
+}
+
+impl<'a> Cursor<&'a mut [u8]> {
+    fn write(&mut self, bytes: &[u8]) {
+        let next = self.pos + bytes.len();
+        self.buf[self.pos..next].copy_from_slice(bytes);
+        self.pos = next;
+    }
+
+    fn write_tagged(&mut self, bytes: &[u8]) {
+        bytes.tag().serialize(self);
+        self.write(bytes);
+    }
+}
+
+impl<'a> Cursor<&'a [u8]> {
+    fn read(&mut self, count: usize) -> &'a [u8] {
+        let read = &self.buf[self.pos..self.pos + count];
+        self.pos += count;
+        read
+    }
+
+    fn read_tagged(&mut self) -> &'a [u8] {
+        let count = u32::deserialize(self);
+        self.read(count as usize)
+    }
+}
+
+trait Tagged<'a> {
+    fn tag(&self) -> u32;
+}
+
+impl<T> Tagged<'_> for [T] {
+    fn tag(&self) -> u32 {
+        self.len() as u32
+    }
+}
+
+impl Tagged<'_> for str {
+    fn tag(&self) -> u32 {
+        self.as_bytes().tag()
+    }
+}
+
+/// Serializes data structure into byte slice.
+pub trait Serialize {
+    /// Write self into buffer.
+    ///
+    /// # Panics
+    ///
+    /// If [`size`](Serialize::size) is incorrectly implemented, buffer can be
+    /// too small to fit all bytes.
+    fn serialize(&self, buffer: &mut Cursor<&mut [u8]>);
+
+    /// Calculate resulting size of serialized data structure.
+    ///
+    /// Resulting value will be used to allocate buffer, so incorrect implementation
+    /// will either result in panic after out of bounds access or wasted space.
+    fn size(&self) -> usize;
+
+    /// Same as [`size`](Serialize::size), but with space for slice length (now [`u32`])
+    #[doc(hidden)]
+    fn size_tagged(&self) -> usize {
+        size_of::<u32>() + self.size()
+    }
+}
+
+/// Deserializes byte slice into desired data structure.
+pub trait Deserialize<'a> {
+    /// Read bytes from buffer and interpret as desired.
+    ///
+    /// # Panics
+    ///
+    /// If provided buffer is messed up (i.e. it isn't filled by corresponding
+    /// [`Serialize`] implementation).
+    ///
+    /// # Safety
+    ///
+    /// This function isn't marked as unsafe, so in no circumstances can
+    /// implementations count on provided buffer to be well formed. Panic instead.
+    fn deserialize(buffer: &mut Cursor<&'a [u8]>) -> Self;
+}
+
+macro_rules! primitive {
+    ($($type:ty),+ $(,)?) => {
+        $(
+        impl Serialize for $type {
+            fn serialize(&self, buffer: &mut Cursor<&mut [u8]>) {
+                self.to_ne_bytes().serialize(buffer)
+            }
+
+            fn size(&self) -> usize {
+                size_of::<$type>()
+            }
+        }
+
+        impl Deserialize<'_> for $type {
+            fn deserialize(buffer: &mut Cursor<&[u8]>) -> Self {
+                let array = buffer.read(size_of::<$type>()).try_into().expect("slice is correctly sized");
+                Self::from_ne_bytes(array)
+            }
+        }
+        )+
+    };
+}
+
+primitive!(u8, u16, u32, u64, u128, usize);
+primitive!(i8, i16, i32, i64, i128, isize);
+primitive!(f32, f64);
+
+impl Serialize for [u8] {
+    fn serialize(&self, buffer: &mut Cursor<&mut [u8]>) {
+        buffer.write(self);
+    }
+
+    fn size(&self) -> usize {
+        size_of_val(self)
+    }
+}
+
+impl Serialize for str {
+    fn serialize(&self, buffer: &mut Cursor<&mut [u8]>) {
+        self.as_bytes().serialize(buffer)
+    }
+
+    fn size(&self) -> usize {
+        size_of_val(self)
+    }
+}
+
+impl<T: Serialize> Serialize for Vec2<T> {
+    fn serialize(&self, buffer: &mut Cursor<&mut [u8]>) {
+        self.x.serialize(buffer);
+        self.y.serialize(buffer);
+    }
+
+    fn size(&self) -> usize {
+        2 * size_of::<T>()
+    }
+}
+
+impl<T: Serialize> Serialize for Box<[T]> {
+    fn serialize(&self, buffer: &mut Cursor<&mut [u8]>) {
+        self.tag().serialize(buffer);
+        for item in self.iter() {
+            item.serialize(buffer);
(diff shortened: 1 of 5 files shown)


common/src/ipc/serde.rs
```diff
@@ -0,0 +1,631 @@
...
+            }
+
+            fn size(&self) -> usize {
+                size_of::<$type>()
+            }
+        }
+
+        impl Deserialize<'_> for $type {
+            fn deserialize(buffer: &mut Cursor<&[u8]>) -> Self {
+                let array = buffer.read(size_of::<$type>()).try_into().expect("slice is correctly sized");
```
LGFae, line 135: I believe this panic message is wrong. If the slice is incorrectly sized, it will panic in `read`, not here. This will fail if the conversion from `try_into` fails.

common/src/mmap.rs
```diff
@@ -20,8 +21,7 @@ use rustix::shm::ShmOFlags;
 #[derive(Debug)]
 pub struct Mmap {
     fd: OwnedFd,
-    ptr: NonNull<std::ffi::c_void>,
-    len: usize,
+    buffer: NonNull<[u8]>,
```
LGFae, line 24: Doesn't this add extra indirection? We have a pointer to an array, which is a pointer and a len. There's also no point in using NonNull here, since [u8] can, by Rust's rules, never be null.

If you want to keep the indirection it's better to use a `Box<[u8]>`. Otherwise, you can keep the code as it was before, maybe just with a pointer to `u8` directly, rather than `c_void`: `NonNull<u8>`.

LGFae (commented):
I may take me a bit longer than usual because I nearing the end of my final semester as an undergrad, so I apologize for that.

common/src/mmap.rs
LGFae: Also, I believe just keeping the `NonNull<u8>` directly may be better simply because you will end up manipulating the buffer's core components anyways later in the code.

common/src/ipc/serde.rs
rkuklik: `TryInto` will fail if `slice.len()` isn't equal to number of elements in requested array. This is why I expect slice to have correct size.

common/src/mmap.rs
rkuklik: `NonNull<T>` is defined as `*mut T` with rustc attribute for layout optimizations, so there is a reason. 

> [u8] can, by Rust's rules, never be null.

I don't understand. `[u8]`, by itself, can never exist. It must be behind a pointer.
So `&[u8]`, `*mut [u8]` and friends are basically just `struct Pointer { addr: usize, len: usize }`. I think I misunderstood.

> If you want to keep the indirection it's better to use a `Box<[u8]>`. Otherwise, you can keep the code as it was before, maybe just with a pointer to `u8` directly, rather than `c_void`: `NonNull<u8>`.

It is not. The buffer was provided by `mmap`, so on `Drop`, the `Box` would take the pointer, go to system allocator, try to deallocate and probably segfault.

I think `NonNull<[u8]>` conveys that `Mmap` is guaranteed to own slice of bytes but is unable to use `&mut [u8]` because it couldn't assign a lifetime. `'static` is incorrect, because the lifetime is however long `Mmap` will live.

Given that `Mmap`s purpose is just to be a glorified buffer, it is better to use `[u8]` as the type pointed to.

rkuklik:
> I may take me a bit longer than usual because I nearing the end of my final semester as an undergrad, so I apologize for that.

Of course, take your time. On that note, next two weeks I will also be less available, so no pressure.

common/src/ipc/serde.rs
LGFae: You are right, my mistake.

common/src/mmap.rs
LGFae: You are right on all accounts. I am the one who got confused, my apologies.

I also agree with your arguments on why to use `[u8]` rather than `u8`.

common/src/ipc/serde.rs
```diff
@@ -0,0 +1,631 @@
...
+use super::types2::ImageRequest;
+use super::types2::Info;
+use super::types2::PixelFormat;
+use super::types2::Scale;
+use super::types2::Transition;
+use super::types2::TransitionType;
+use super::types2::Vec2;
+
+/// Type for managing access in [`Serialize`] and [`Deserialize`]
+pub struct Cursor<T> {
```
itsjunetime, line 22: This seems nearly identical to `std::io::Cursor` (just with a usize vs a u64) - why don't we just use that?

common/src/mmap.rs
```diff
@@ -33,18 +33,17 @@ impl Mmap {
     #[must_use]
     pub fn create(len: usize) -> Self {
```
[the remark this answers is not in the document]
rkuklik: No idea. Seems nice though, I would love to use it.

common/src/ipc/transmit.rs
```diff
@@ -181,9 +191,10 @@ impl From<RawMsg> for Answer {
 }
 // TODO: end remove ugly mess block
 
-macro_rules! code {
+macro_rules! msg {
     ($($name:ident $num:literal),* $(,)?) => {
-        pub enum Code {
+        #[derive(Copy, Clone, PartialEq, Eq)]
```
itsjunetime, line 196: I don't know if it would be worth it for another dependency, but the `strum` crate has derive macros that can do this conversion between a repr for you (`FromRepr`) - also, if you mark `Code` as `#[repr(u64)]`, you can just do e.g. `Code::ReqPing as u64` for free without needing the `into` fn.

common/src/ipc/transmit.rs
```diff
@@ -288,6 +305,131 @@ impl<T> IpcSocket<T> {
...
+}
+
+impl<'a> From<&'a IpcMessage> for Request<'a> {
+    fn from(value: &'a IpcMessage) -> Self {
+        match value.code {
+            Code::ReqPing => Self::Ping,
+            Code::ReqQuery => Self::Query,
+            Code::ReqKill => Self::Kill,
+            Code::ReqClear => {
+                let mmap = value.shm.as_ref().expect("clear request must contain data");
```
itsjunetime, line 354: Shouldn't `IPCMessage` be an enum to encode these invariants (of `Clear` and `Img` requiring `shm` to be `Some`?

common/src/ipc/serde.rs
rkuklik: I looked at it, but didn't see a reason to use it. It implements `BufRead` and `Write`, but I would want to panic anyway with the way I allocate memory (with `size` on `Serialize`). I would also implement extension traits to ease tagged slice writing, so I just wrote couple of lines and left it at that.

`io::Cursor` was the first thing I looked at, but disliked being unable to directly access it.

common/src/ipc/transmit.rs
```diff
@@ -288,6 +305,131 @@ impl<T> IpcSocket<T> {
...
+        match value.code {
+            Code::ResOk => Self::Ok,
+            Code::ResAwait => Self::Ping(false),
+            Code::ResConfigured => Self::Ping(true),
+            Code::ResInfo => {
+                let mmap = value.shm.as_ref().expect("info request must contain data");
+                let info = Box::<[Info]>::deserialize(&mut Cursor::new(mmap.slice()));
+                Self::Info(info)
+            }
+            _ => unreachable!("`Response` builder reached invalid state"),
```
itsjunetime, line 406: Why is this state invalid? imo I feel like it should at least be mentioned in a comment, if not the message itself, why those are invalid states (and maybe the `unreachable!` could explicitly say what state it's in right now for debugging purposes). If these invalid states could be prevented the type system somehow, though, that would obvs be best.

common/src/ipc/transmit.rs
```diff
@@ -288,6 +305,131 @@ impl<T> IpcSocket<T> {
...
+                let mmap = value.shm.as_ref().expect("clear request must contain data");
+                let clear = ClearRequest::deserialize(&mut Cursor::new(mmap.slice()));
+                Self::Clear(clear)
+            }
+            Code::ReqImg => {
+                let mmap = value.shm.as_ref().expect("image request must contain data");
+                let image = ImageRequest::deserialize(&mut Cursor::new(mmap.slice()));
+                Self::Img(image)
+            }
+            _ => unreachable!("`Request` builder reached invalid state"),
```
itsjunetime, line 363: Same question as below about the `unreachable!` with the invalid states

common/src/ipc/serde.rs
```diff
@@ -0,0 +1,631 @@
...
+}
+
+impl Tagged for str {
+    fn tag(&self) -> u32 {
+        self.as_bytes().tag()
+    }
+}
+
+/// Serializes data structure into byte slice.
+pub trait Serialize {
```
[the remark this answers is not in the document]
rkuklik: I wouldn't be against it, @LGFae expressed that he wouldn't want to include it. See [comment](https://github.com/LGFae/swww/pull/331#issuecomment-2184159415).

common/src/ipc/transmit.rs
rkuklik: I have to convert back from `u64`, which (to my knowledge) cannot be done with `num as Code`. Also, owner of this project expressed his desire not to add dependencies.

common/src/ipc/transmit.rs
rkuklik: It should, I think I left a todo comment there, just was too lazy to implement it (yet).

common/src/ipc/transmit.rs
rkuklik: The message consists of code and optional payload. When I read via socket and shove it in `IpcMessage`, I then let this message be interpreted by calling code. If you get a message and expect it to be a `Response`, getting `Request` breaks this expectation. I didn't find elegant way to do this, but at least I will try to improve the diagnostics.

common/src/ipc/serde.rs
itsjunetime: Sorry, I don't know if I'm following - `io::Cursor` has `get_ref()` and `get_ref_mut()` methods to get a reference to the underlying data, so you can directly access it.

And are you saying that when writing a wrong amount of data to the `Cursor` you'd rather panic than get back information about how much was written? That would make sense as a rational, I guess, but feels like less ideal for a user in the long run.

common/src/ipc/serde.rs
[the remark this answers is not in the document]
itsjunetime: Huh. @LGFae could we revisit that? Serde does, by default, copy a lot of data around, but utilizing slices and `#[serde(borrow)]` should allow you to avoid all of that. Or is it something else that's an issue?

common/src/ipc/transmit.rs
rkuklik: See above. Thank you very much for reviewing, I would wait for owners opinion on additional dependencies and the get to work. If you would want to add some changes to this PR yourself, can you, or do I have to add you somehow?

common/src/ipc/serde.rs
rkuklik: If I read correctly, `get_*` gets underlying value, not remaining. There is `remaining_slice`, which is nightly for some reason. So I just implemented bare minimum of what would be required.

And with the panics, given that (currently) reading and writing is constrained to `common` crate, I wanted to implement fuzzing for correctness. So I just opted for panics.

common/src/ipc/serde.rs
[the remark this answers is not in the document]
rkuklik: I think it would simplify the codebase a great deal. +1 from me.

common/src/ipc/serde.rs
LGFae: > And are you saying that when writing a wrong amount of data to the Cursor you'd rather panic than get back information about how much was written? That would make sense as a rational, I guess, but feels like less ideal for a user in the long run.

I disagree with the last part; I think it makes *more* sense for the user to panic. This isn't a recoverable error anyway: if our serialization/deserialization fails, something fundamentally wrong has happened (for example, maybe a different program is trying to write something malicious to the `swww` socket). It also makes for a much easier-to-debug error, since it's just a panic.

common/src/ipc/transmit.rs
LGFae: Like I said before, I specially don't like pulling in dependencies just to use a single function or feature, a single time. In fact, if I was writing this PR, I wouldn't even have used a macro at all. I would have instead written everything manually, though I can see the benefits of the macro @rkuklik wrote.

common/src/ipc/transmit.rs
LGFae: If you can at least build a better error message, I think that would greatly improve debugging and figuring out what went wrong when people report issues.

common/src/ipc/transmit.rs
LGFae: I'd also like to thank you for you input, @itsjunetime. I've had a hard time reviewing this thoroughly, so your help is much appreciated.

common/src/ipc/serde.rs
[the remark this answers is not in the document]
rkuklik: As June wrote, there is a [crate](https://lib.rs/crates/postcard), which has a similar format to my current one. Usually, you can just derive desired traits, so no major refactoring needed.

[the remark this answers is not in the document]
LGFae: Ah, my bad, I thought `postcard` was like, a standard format like `protobuf` or something. It seems to be a lot more flexible than that.

Sure, we can try it out! But I do think we should do it in a separate PR, because I feel like it will be a pretty big change.

rkuklik:
Hello, sorry for the delay. Given that my previous PR broke swww, I would close this PR and reopen it after I fix the breakage.

I have looked into `postcard` and must say, +1 for me. I consider dependencies one of the biggest wins with Rust, and would like to introduce `serde` and `postcard` to `common` crate. `serde` is goto solution for the ecosystem as a whole and `postcard` seems decent to me.

What do you think?

LGFae:
Hey there. I am going to slowly get back into this over the course of the following weeks.

I've thought a lot about using `serde` and dependencies in general after your and @itsjunetime's comments. There's really just two problems I have with dependencies:

1. they increase compilation time -- this isn't *super* troublesome, because most users will only pay the price for this once. It does make development more annoying, but not by much since we can use debug builds.
2. they increase resource consumption in the `daemon`. This is the big one. Because the `daemon` will run "forever", I would really like for it to be as efficient as reasonably possible. For example, right now, after setting an image, we unmap the underlying memory, such that `RSS` reports that just around 2.5MB are being used. I would really like to keep this behavior, and it's not going to be possible with higher-level libraries.

So, in conclusion, if you think using `serde` and `postcard` is possible while retaining the ability to unmap the underlying image bytes' memory, let's go for it.

rkuklik:
Hello. I will close this and get back to it when I fix the regression with socket. As to the two points:


> 1. they increase compilation time -- this isn't _super_ troublesome, because most users will only pay the price for this once. It does make development more annoying, but not by much since we can use debug builds.

About speed of compilation, that depends. Cargo can concurrently compile multiple crates, so the biggest slowdown is in downloading them (with the now resolver, shouldn't be a big deal). When recompiling, you only recompile your crates, not the whole dependency graph, so it shouldn't impact development (can speed it up even), with the exception of link-time optimization (which can be turned off) and macro expansion (which shouldn't be an issue, only ones we want to be using are for `serde`).

> 2. they increase resource consumption in the `daemon`. This is the big one. Because the `daemon` will run "forever", I would really like for it to be as efficient as reasonably possible. For example, right now, after setting a static image, we unmap the underlying memory, such that `RSS` reports that just around 2.5MB are being used. I would really like to keep this behavior, and it's not going to be possible with higher-level libraries.

I am not sure, why should they increase resource usage? You can stumble upon a poorly written crate, sure, but I would say using widely used and well maintained library can improve correctness and efficiency of your code. Big selling point of Rust is the ability to use high level abstractions without the penalties usually associated with them (`Iterator` would be a great example).

I think it is stupid to use crates like [`console_error_panic_hook`](https://github.com/rustwasm/console_error_panic_hook/blob/master/src/lib.rs) which is about 20 lines of trivial
code, but for complex tasks it is beneficial to lean into other people's work.

> So, in conclusion, if you think using `serde` and `postcard` is possible while retaining the ability to unmap the underlying image bytes' memory, let's go for it.

Sure. I will fix the regression and get to it. Thank you and @itsjunetime for your work.

PR #340, opened 2024-07-10
add swww_randomize_multi.sh example script

cplir-c:
The swww_randomize_multi script chooses a different random image for each display, instead of setting the same image for all displays.


LGFae:
Thanks!

PR #372, opened 2024-10-24
Add image filter to cache

kerty0:
I primary use pixel art wallpapers, so the default filter for all restored images was annoying.
I modified cache so it stores filter and path to image separated by '\n' (I couldn't think of anything better). It can work with old version of cache if it doesn't contain '\n' using default filter.
I also fixed a crash that occurred when using a solid color on my system (EndevourOS + Hyprland): the drawable surface required RGBA image but was only provided with RGB.
I'm still learning Rust and English is not my native language, so sorry if anything is wrong.

### DIFF
--- a/client/src/cli.rs
+++ b/client/src/cli.rs
@@ -83,8 +83,7 @@ impl Display for Filter {
             Self::CatmullRom => "CatmullRom",
             Self::Mitchell => "Mitchell",
             Self::Lanczos3 => "Lanczos3",
-        }
-        .to_string();
+        };
         write!(f, "{}", str)
     }
 }
--- a/client/src/cli.rs
+++ b/client/src/cli.rs
@@ -1,6 +1,7 @@
 /// Note: this file only has basic declarations and some definitions in order to be possible to
 /// import it in the build script, to automate shell completion
 use clap::{Parser, ValueEnum};
+use std::fmt::Display;
 use std::path::PathBuf;
 
 fn from_hex(hex: &str) -> Result<[u8; 3], String> {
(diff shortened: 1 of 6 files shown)


LGFae:
> It can work with old version of cache if it doesn't contain '\n'

We don't need this. Instead, increase the cargo version from masterV2 to masterV3. The cache will won't consider files with older versions.

kerty0:
Now get_previous_image_path returns error to the old version of cache and users will be forced to set the wallpaper again to update cache.

client/src/cli.rs
```diff
@@ -74,6 +75,20 @@ impl std::str::FromStr for Filter {
...
+impl Display for Filter {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let str = match self {
+            Self::Nearest => "Nearest",
+            Self::Bilinear => "Bilinear",
+            Self::CatmullRom => "CatmullRom",
+            Self::Mitchell => "Mitchell",
+            Self::Lanczos3 => "Lanczos3",
+        }
+        .to_string();
```
LGFae, line 87: You shouldn't need the to_string call here.

common/src/cache.rs
```diff
@@ -64,28 +64,35 @@ pub fn load_animation_frames(
...
 
-pub fn get_previous_image_path(output_name: &str) -> io::Result<String> {
+pub fn get_previous_image_path(output_name: &str) -> io::Result<(String, String)> {
     let mut filepath = cache_dir()?;
     clean_previous_verions(&filepath);
 
     filepath.push(output_name);
     if !filepath.is_file() {
-        return Ok("".to_string());
+        return Ok(("Lanczos3".to_string(), "".to_string()));
```
LGFae, line 73: I don't really get what this gives us. Is it just to improve the error messages?

LGFae (commented):
Very nice. Just two small things and we can merge this.

kerty0:
How swww restore don't stop half done if one of the outputs don't have cache file and print name of the output for all errors

common/src/cache.rs
kerty0: Yes, and it makes it easy to force swww-daemon ignore missing cache files.

LGFae:
Thanks!

Issue #166, opened 2023-08-28
Common resize options


anufrievroman:
Hi, I am a maintainer of [waypaper](https://github.com/anufrievroman/waypaper) app, that uses `swww` as one of the backends for wallpaper setting. I have one suggestion:

Classically, wallpaper fill modes are either `fill`, `fit`, `center`, `stretch`, or `tile`, which are all different things. However, `swww --resize` only supports `no`, `crop`, and `fit`. So, I guess `no` is `center` and `crop` is `fill`, but `stretch` and `tile` are not present. I'd suggest implementing these modes as well, just to be more uniform with other wallpaper tools. Thank you.

Visual-Dawg:
Also adding the option for `no resize` to scale images to the screen size if they are larger than the screen, but otherwise leave them as is, would be really nice :)

Sx1Fl0NdqlkM6N4s2pswXKUI0k9FKTaOckLbNXu:
given the amount of smartphone wallpapers/photos, having few options to i.e. repeat one (or multiple?) wallpaper, horizontally only, aligned from center, fit to screen vertically, would be quite useful.

JasonBoyett:
There is a stretch option in the current version. However it hasn't been added to an actual release yet so in order to get it you will have to clone the main branch and build from source. 

Just in case here's a quick walk through on how to do that.

> [!NOTE]
> make sure you have [cargo](https://github.com/rust-lang/cargo)


```bash
## clone the repo
git clone https://github.com/LGFae/swww

## cd into the repo
cd swww

## run cargo build
cargo build

## copy the binary for the client and daemon to your bin
cp target/debug/swww-daemon /usr/bin/swww-daemon
cp target/debug/swww /usr/bin/swww
```

Now you should be able to use the stretch feature.

PR #373, opened 2024-10-26
Add stretch to resize options

rexept:
I added stretch to the --resize options. The function is almost identical to the crop function with just a minor change. I have tested it locally and it seems to be working correctly.

Something similar to this was mentioned in issue #166


LGFae:
From my tests it also appears to be working. Thanks! Can you just run `cargo fmt` before we merge?

rexept:
Maybe test again because I realized I was testing on this latest commit I pushed. I has a typo fix.

LGFae:
Seems fine. Thanks again!

PR #375, opened 2024-10-31
Fix memory leak and stopped animation when connecting/disconnecting monitor

kerty0:
1. Memory leak when connecting/disconnecting monitor (happens on KDE Plasma and Hyprland): if I understand it correctly, Compositor sends LayerSurface closed event which removes Wallpaper belonging to closed surface from Daemon.wallpapers, later Compositor sends WL_REGISTRY global_remove which tries to delete Wallpaper belonging to closed output and stop it animation by searching Daemon.wallpapers by output_name, but fails because Wallpaper has already been removed. This leads to not deleted TransitionAnimator and ImageAnimator, which stops Wallpaper from being dropped and starts memory leak in buffer leading to swww-daemon termination.
Because I don't know right way to handle this I just duplicated code from global_remove event to closed event so both now stop animations.
2. Stop of animation on primary monitor when connecting/disconnecting second one (only on Hyprland): when disconnecting second, monitor Hyprland sends wl_output done event for primary monitor, and when connecting second monitor Hyprland sends sequence of events with done event in the end for both monitors. wl_output done event stops animations so connecting/disconnecting second monitor stops wallpaper animation on first one.
I don't know right way to handle this either so I just removed stop_animations from wl_output done event.

### DIFF
--- a/common/src/ipc/types.rs
+++ b/common/src/ipc/types.rs
@@ -86,6 +86,10 @@ impl BgImg {
             Self::Img(s) => 4 + s.len()
         }
     }
+
+    pub fn is_set(&self) -> bool {
+        matches!(self, Self::Img(_))
+    }
 }
 
 impl fmt::Display for BgImg {
@@ -147,7 +151,7 @@ impl PixelFormat {
     }
 }
 
-#[derive(Clone, Copy, Debug, PartialEq)]
+#[derive(Clone, Copy, Debug)]
 pub enum Scale {
     Whole(NonZeroI32),
     Fractional(NonZeroI32),
@@ -180,6 +184,18 @@ impl Scale {
                 (width, height)
             }
         }
+    }
+}
+
+impl PartialEq for Scale {
+    fn eq(&self, other: &Self) -> bool {
+        (match self {
+            Self::Whole(i) => i.get() * 120,
+            Self::Fractional(f) => f.get(),
+        }) == (match other {
+            Self::Whole(i) => i.get() * 120,
+            Self::Fractional(f) => f.get(),
+        })
     }
 }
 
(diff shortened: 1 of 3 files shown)


LGFae:
I will have to test this manually (might take a little bit), but I am pretty sure solution 2 is wrong because animations will continue to play if you change the output's configuration. Looking at `Wallpaper::commit_surface_changes`, it appears to return a boolean indicating whether or not we have to stop the animation. I think you can just use it instead. This is what we do in the `preffered_scale` function:

```rust
if wallpaper
       .borrow_mut()
       .commit_surface_changes(self.use_cache)
{
        self.stop_animations(&[wallpaper.clone()]);
}
```

Solution 1, on the other hand, appears fine.

kerty0:
I thinking about adding to `Daemon` something like this:
``` rust
fn reconfigure_wallpaper(&self, output_name: String) {
    std::thread::Builder::new()
        .name("wallpaper reconfigurator".to_string())
        .stack_size(1 << 14)
        .spawn(move || {
            if let Err(e) = || -> io::Result<()> {
                std::process::Command::new("swaww")
                    .arg("restore")
                    .arg(format!("--outputs={}", output_name))
                    .spawn()?
                    .wait()?;
                Ok(())
            }() {
                warn!("failed to reconfigure wallpaper: {e}");
            }
        })
        .unwrap();
}
```
And calling it when scaling or resolution of monitor changes so wallpaper always displays correctly. What do you think?

LGFae:
We are already doing this though. When we receive an `output done` event we call `wallpaper::commit_surface_changes`,  which calls `common::cache::load` in a different thread.

There is a problem in that if the daemon is called with the `--no-cache` flag, then it will never reload the image, even when the output is reconfigured. Maybe that's the problem you would like to solve?

Issue #386, opened 2024-12-29
Wallpaper changing animations do not work on vertical monitor.


GeodeArc:
On my vertical monitor (DP-2), there is no animation when switching between wallpapers, and it just cuts.

Things I have tried:
- All animation types
- Isolating the vertical monitors output (disabling all other displays)
- Using the latest swww-git release

https://github.com/user-attachments/assets/52ac1217-c265-43f7-a8b3-da9bac150859

Any help/fixes are appreciated!

martobtw:
I'm not sure in your case but this works for me.
https://github.com/JaKooLit/Hyprland-Dots/blob/main/config/hypr/UserScripts/WallpaperSelect.sh

GeodeArc:
Interesting, I have been able to replicate this on a virtual machine, and also my laptop. ill play around a bit more and see what i can do.

I also have a very similar (but less developed) script that changes wallpapers like the one you provided, and sadly experience the issue.

Just so I know im not going crazy, do you mind trying to replicate this setup and testing?
1. Install swww-git (the regular swww package has another vertical monitor issue that hasnt yet been merged from git).
2. Change a monitors transform (rotation) to 1 or 3 (aka vertical) in hyprland config.
3. Change wallpaper and check animation displayed (if any).

Sometimes I see the start of an animation, maybe for half a second, but then it immediately cuts.

martobtw:
Mine is already like that. Also had that problem with the square wp so i updated to git ver. My animation when switching is working. I will test it out again tomorrow because I am in bed rn.

Ant01n3:
Hello! Same problem here with the last git version and Hyprland 46.2. It works very well on the main monitor in landscape, but not on the secondary monitor in portrait mode, no animation or only the start of it. Moreover, the daemon prints "failed to load cache: there is already another swww process running" several times only when changing the image on the vertical monitor. It also seems that the CPU is more active as long as I don't send another image to the landscape monitor. I can do tests if needed.

Nynxz:
narrowed it down to [This Commit (56928f7c0e55a321ad5a83a039e0e0dc3eeac35b)](https://github.com/LGFae/swww/commit/56928f7c0e55a321ad5a83a039e0e0dc3eeac35b)  causing the issue

reverting it and building reset to HEAD seems to work.

will try to figure out a fix rather than just removing entirely before pushing a merge request.. 
haven't written rust, if anyone gets to it before me no worries.

GeodeArc:
[...]

PR #389, opened 2025-01-17
fix vertical monitor transitions & 'failed to load cache' errors

Nynxz:
Fixed Issues: 
 - Transitions on vertical monitor's were having issues
 - `[WARN]  failed to load cache: there is already another swww process running` errors on wallpaper changes
 
 Discussion: [Wallpaper changing animations do not work on vertical monitor
](https://github.com/LGFae/swww/issues/386)

have not written rust, 
I am not sure if this causes issues with https://github.com/LGFae/swww/commit/56928f7c0e55a321ad5a83a039e0e0dc3eeac35b

### DIFF
--- a/daemon/src/wallpaper.rs
+++ b/daemon/src/wallpaper.rs
@@ -252,13 +252,7 @@ impl Wallpaper {
         use wl_output::transform;
         let inner = &mut self.inner;
         let staging = &self.inner_staging;
-
-        if (inner.name != staging.name && use_cache)
-            || (self.img.is_set()
-                && (inner.scale_factor != staging.scale_factor
-                    || inner.width != staging.width
-                    || inner.height != staging.height))
-        {
+        if inner.name != staging.name && use_cache {
             let name = staging.name.clone().unwrap_or("".to_string());
             std::thread::Builder::new()
                 .name("cache loader".to_string())


LGFae:
I believe the problem with this is that it wouldn't reload the image if the output's configuration changed. That was the original reason for that commit.

LGFae:
I also do not understand how this could possibly be the solution to #386. There seems to be a more fundamental issue we have to address.

Nynxz:
https://github.com/user-attachments/assets/da9f0e5f-f948-469e-9e09-abf96dae35ad

Not sure, the issue did not exist before https://github.com/LGFae/swww/commit/56928f7c0e55a321ad5a83a039e0e0dc3eeac35b
I feel it would be better to remove the problem code and try fix what was intended by https://github.com/LGFae/swww/commit/56928f7c0e55a321ad5a83a039e0e0dc3eeac35b (refreshing when monitor input changes) a different way 
over trying to fix the issue https://github.com/LGFae/swww/commit/56928f7c0e55a321ad5a83a039e0e0dc3eeac35b caused. I.e listening to actual wayland events, rather than constantly polling for whether the 'surface has changed', which is what it seems like its doing currently, and causing a memory leak. 

there is also a memory leak if transitions get overwritten midway through (changing the wallpaper too quickly)
 but that is a separate issue to these ones.

GeodeArc:
It seems to be working now (after upgrading to Hyprland 47), so I closed the original. Unfortunately so many things broke in hyprland 47 for me that I will be switching back to 46 for now, meaning I have to live with the animations.

LGFae:
Alright, we may have the proper fix for this in PR #439, actually.

PR #391, opened 2025-01-25
Make example scripts more portable

rubiefawn:
This is an attempt to make `swww_randomize.sh` and `swww_randomize_multi.sh` more portable by replacing the `$RANDOM` `bash`-ism with more portable way of randomizing the files.

This also removes `swww_randomize_multi.sh`'s check to make sure it is not already running, since it uses `$BASHPID`. In my subjective opinion, such logic is not needed in the first place, since these scripts are likely to be run once by the compositor automatically rather than manually invoked multiple times. That said, if it is required, I will write a portable solution that doesn't use `$BASHPID` and apply that solution in `swww_randomize.sh` as well, which currently lacks this check.

### DIFF
--- a/example_scripts/swww_randomize.sh
+++ b/example_scripts/swww_randomize.sh
@@ -1,32 +1,28 @@
-#!/bin/bash
+#!/bin/sh
+# Changes the wallpaper to a randomly chosen image in a given directory
+# at a set interval.
 
-# This script will randomly go through the files of a directory, setting it
-# up as the wallpaper at regular intervals
-#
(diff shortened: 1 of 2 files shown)


JS-CodeGamer:
Just a suggestion, and I am not sure about portability here but why not use `shuf` or `sort -R` instead of using `... "number:file" | sort -n ...` ?

filip-rs (approved):
LGTM, both scripts work fine on my machine (Arch + hyprland). Good change, could be merged.

LGFae:
My bad, I had forgotten to merge this. Thanks!

Issue #393, opened 2025-01-26
Left click on the wallpaper in KDE6 restores the KDE wallpaper


MaxandreOgeret:
As explained, the wallpaper is set with swww (I have 3 screens so 3 wallpapers set with swww).
Setting the wallpaper works fine.

But if I left click on the desktop, the default KDE Plasma wallpaper is restored. Restarting the daemon and re-applying the wallpaper restores the sww wallpapers.

Any way around that ?

Thanks !

 - swww 0.9.5-masterV3
 - plasmashell 6.2.5

Ghoul072:
Same issue here

plasmashell 6.3.0
swww 0.9.5

i've tried adding a fully transparent color bg and removing WallpaperPlugin from plasmashellrc as well. The swww wallpaper still seems to get pushed behind a solid black background.

`kquitapp6 plasmashell`  reveals that swww daemon was running fine all along behind the black screen. Restarting plasmashell brings the black background on top of the swww wallpaper again. There might be a way to keep swww on top of plasmashell but i'm not familiar

hypnotics-dev:
I am also seeing this issue with swww 0.9.5 and plasmashell 6.3. Is there any way of collecting any more useful data to help with the debugging process?

LGFae:
The best way would probably be making a debug build and running that.

cristiancolosimo:
Hi, 

I have the same problem for the left click, but sometime also for the right click.

For fix (hack)  it, in first place i think to restart every time the daemon before change the display, and set the kde plasma wallpaper same as the swww so when i click i show the same wallpaper and i don't notice the "bug", but sometime when i restart swww it show a little animation of growing and isn't beautiful.

My script for set KDE wallpaper same of swww.
The script is triggered by a systemd unit(user) with a timer every 3 minute.
swww-daemon is also managed by a custom systemd unit (user)
```bash
#!/bin/bash
IMAGE=$(find /home/USERNAME/Immagini/wallpaper  -maxdepth 1 -type f | shuf -n 1)
TRANSITION=("any"  "wave" "outer"  "wipe" )
RI=$((RANDOM % ${#TRANSITION[@]}))
RI="${TRANSITION[$RI]}"
swww img --transition-type $RI $IMAGE
qdbus org.kde.plasmashell /PlasmaShell org.kde.PlasmaShell.evaluateScript "
    desktops().forEach((d) => {
        d.currentConfigGroup = [
            'Wallpaper',
            'org.kde.image',
            'General',
[...]

PR #400, opened 2025-03-01
wallpaper: change layer to BOTTOM for proper working in KDE6

cristiancolosimo:
(sorry, it's first pull request, I don't know what I am doing)

Relative to the Issue #393

### DIFF
--- a/daemon/src/wallpaper.rs
+++ b/daemon/src/wallpaper.rs
@@ -107,7 +107,7 @@ impl Wallpaper {
             layer_surface,
             wl_surface,
             Some(output),
-            zwlr_layer_shell_v1::layer::BACKGROUND,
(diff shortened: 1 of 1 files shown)


LGFae:
I do not think this is the proper fix. The Background layer is so called because it is supposed to be used for backgrounds. Setting it to Bottom could potentially cause issue with things like status bars.

cristiancolosimo:
Thanks for reply,

Yeah, I thought so too, but on kde I'm not having any problem, I have no way to try it on other environments at the moment. Tomorrow I'll try to test it on gnome. Can you give me some examples of status bars to test, which I install and test with the default configs.

Axlefublr:
yipeeee thanks

PR #410, opened 2025-04-05
Use `waybackend` for our wayland code

LGFae:
I have reimplemented all the wayland logic into a separate crate. This will allow me to reuse that in different projects, and also let us delete large swaths of code in `swww`. A lot of this code was nasty `unsafe` stuff, that we have moved to a different crate that will (hopefully) be better tested.


LGFae:
I was basically taking this for a test run in my desktop and notebook over the last few weeks. I believe I have found no issues. The problem is that I can't get the nix-test to build correctly in CI.

I am thinking of just deleting the test altogether. It's not nice having a test for a system I don't use and cannot easily verify myself.

Issue #425, opened 2025-05-26
Incorrect query result on 0.10.0


DagHemberg:
![Image](https://github.com/user-attachments/assets/5a07f308-3647-45fc-b2bb-45979fe0fd29)

Currently running Fedora 42 with Hyprland 0.49.0, swww version 0.10.0 

```
$ hyprctl monitors | rg "(eDP| at |scale)"
Monitor eDP-1 (ID 0):
	2560x1600@120.00100 at 0x0
	scale: 1.60

$ swww query
eDP-1: 1280x800, scale: 2, currently displaying: image: /home/dag/.cache/timewall/wallpapers/1d937151ce86250c/4.png
```

Currently I'm not running any special `monitor = ...` configs for hyprland, but adding one with the correct values hardcoded didn't seem to change anything regarding this issue.

Checking the logs for my `swww-daemon` service gives this (I think the warns are a separate issue though? Including them as well just in case):
```
$ journalctl --user -u swww -f
maj 26 15:03:59 laptop swww-daemon[42236]:     278200ms [WARN]  Received PreferredBufferTransform. We currently ignore those
maj 26 15:03:59 laptop swww-daemon[42236]:     278234ms [WARN]  Received PreferredBufferTransform. We currently ignore those
maj 26 15:03:59 laptop swww-daemon[42236]:     278268ms [WARN]  Received PreferredBufferTransform. We currently ignore those
maj 26 15:03:59 laptop swww-daemon[42236]:     278300ms [WARN]  Received PreferredBufferTransform. We currently ignore those
maj 26 15:03:59 laptop swww-daemon[42236]:     278334ms [WARN]  Received PreferredBufferTransform. We currently ignore those
maj 26 15:03:59 laptop swww-daemon[42236]:     278367ms [WARN]  Received PreferredBufferTransform. We currently ignore those
maj 26 15:03:59 laptop swww-daemon[42236]:     278400ms [WARN]  Received PreferredBufferTransform. We currently ignore those
maj 26 15:03:59 laptop swww-daemon[42236]:     278441ms [WARN]  Received PreferredBufferTransform. We currently ignore those
maj 26 15:03:59 laptop swww-daemon[42236]:     278444ms [ERROR] We failed to find wayland buffer with id: 8. This should be impossible.
maj 26 15:03:59 laptop swww-daemon[42236]:     278444ms [ERROR] We failed to find wayland buffer with id: 14. This should be impossible.
```

LGFae:
[...]

PR #429, opened 2025-05-27
fix fractional scaling

cyuria:
Small mistake which broken fractional scaling after the waybackend split :)


cyuria:
I think this is the same issue as #425

LGFae:
Thanks!

PR #435, opened 2025-05-30
custom namespace support

LGFae:
MISSING: `--all` option for all subcommands, to send things to all namespaces at once.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,15 @@
 
 #### BREAKING
 
+Note: though this are technically breaking changes, they are unlikely to break
+most setups. They are only a problem if you used scripts that relied on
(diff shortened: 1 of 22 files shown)


LGFae:
`--all` option has been implemented. This changes a lot of things so I will be testing it on my machines the following weak. If anyone else would like to test it also, I would appreciate it.

LGFae:
TODO:

- [ ] let users specify multiple namespaces in one command
- [ ] update documentation.

omentic:
@LGFae could we get a new release for this? :^)

(the 0.10.3 release was *just* before this got merged)

LGFae:
Ah, it has been while, hasn't it? I need to do some extra manual testing and also see if I can solve #468 before (because it seems to be pretty simple), and then I will probably do a release either today or maybe tomorrow.

PR #439, opened 2025-06-06
prioritize fractional scale events

LGFae:
We currently let preferred_buffer_scale events overwrite the values we received from the fractional scale protocol. This is incorrect. Wayland is an entirely assynchronous protocol, so we should prioritize the more precise events over the more generic ones.

### DIFF
--- a/common/src/ipc/types.rs
+++ b/common/src/ipc/types.rs
@@ -153,16 +153,31 @@ impl PixelFormat {
 
 #[derive(Clone, Copy, Debug)]
 pub enum Scale {
-    Whole(NonZeroI32),
+    /// sent by wl_output::scale events
+    Output(NonZeroI32),
+    /// sent by wl_surface::preferred_buffer_scale events
+    Preferred(NonZeroI32),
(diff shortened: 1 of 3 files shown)


LGFae:
bad commit... for which issue? Note someone has already pointed at that commit in issue #386, which should also be fixed by this PR.

LOSEARDES77:
Ho sorry i got confused i thought this was for the incorrect scaling issue

LGFae:
Yeah, that's unrelated.

LGFae:
As this was already released, you would need to wait until your distro releases the latest `swww` version (0.10.2), or simply clone this repository and try installing it manually yourself. I cannot comment on how do that as it is specific to every distro.

everyoneexe:
Fix Applied
client/src/imgproc.rs
Line: 452

Removed .fit_into_destination(Some((0.5, 0.5))) — took a different route than you folks, only to realize you patched it more elegantly later on 😅
Anyway, bug's gone. Different scalpel, same surgery.

Issue #48, opened 2022-11-04
Feature Request | Support for svgs


thomascft:

Horus645:
Unlikely to be implemented anytime soon (if at all), as the [image](https://github.com/image-rs/image) crate doesn't seem to have (or plan to have) support for it.

Horus645:
Update: if I manage to finish up #52, doing this wouldn't be so bad. We could use the [resvg](https://crates.io/crates/resvg) crate for it. I won't make any guarantees, however, since using svgs for wallpapers strikes me as very niche. If it makes things too complicated, I probably won't do it.

EdJoPaTo:
SVGs can be rendered pixel perfect for monitors and will always look sharp. Thats my main reason to prefer SVG wallpaper when possible/available. [Gnome also delivers their wallpapers as SVG or WebP.](https://gitlab.gnome.org/GNOME/gnome-backgrounds/-/tree/main/backgrounds)

tralph3:
Just leaving a comment to say that I'd like this to be implemented too. I want to use SVGs as background images since they can be very easily modified with scripts to change their color palette, allowing me to have a set of backgrounds that match my overall colorscheme.

Stetsed:
I don't want to clog up this channel but I too would love to see SVG support.

Horus645:
I was just want to say that reading these comments I have been convinced that it would be nice to have SVG support for multiple users, for a variety of reasons.

Right now I have my hands full with the `sctk 0.17` rewrite, but I think as soon as that is released, and we fix any weird bugs that might arise from it, I will probably look into this.

N0tAI:
If this is something you guys want to have right now I wrote a small shell script to handle it for myself, it should only require librsvg and a shell.

Works on dash and doesn't do anything to simple so I'd reckon almost any sh compatible shell can run it.
```sh
#!/bin/sh

WIDTH='1920px'
HEIGHT='1080px'
OUTPUTS=''

[ -n "$2" ] && WIDTH="$2"
[ -n "$3" ] && HEIGHT="$3"
[ -n "$4" ] && OUTPUTS="$4"

[ -f "$1" ] && rsvg-convert "$1" --width $WIDTH --height $HEIGHT | swww img $([ -n "$OUTPUTS"] && echo "--outputs $OUTPUTS") -
```

Feel free to use it and make it your own

loserMcloser:
[...]

PR #446, opened 2025-06-11
implement support for rendering static SVGs

LGFae:
Solves #48 (finally).


LGFae:
TODO: we need, at a minimum, to fill the color around the svg appropriately, when it does not fill the whole screen.

PR #453, opened 2025-06-23
flake.nix: cleanup

bloxx12:


flake.nix
```diff
@@ -17,37 +16,39 @@
...
-        packages = {
-          swww = rustPlatform.buildRustPackage {
+  }: let
+    inherit (nixpkgs) lib;
+    systems = [
+      "x86_64-linux"
+      "aarch64-linux"
+      "x86_64-darwin"
+      "aarch64-darwin"
+    ];
```
LGFae, line 28: Why are you specifying systems like this? I've gotten issues from people who use `swww` in `x86` systems (not `x86_64`. I don't see what specifying things like this does other than arbitrarily limiting the program's usability.

LGFae (changes requested):
Look, I don't really understand how nix flakes work, so I am going to ask you to explain exactly what you did over here, and why it is a net positive.

bloxx12:
> Look, I don't really understand how nix flakes work, so I am going to ask you to explain exactly what you did over here, and why it is a net positive.

Gladly! The main thing I did was removing `flake-utils`, an abstraction library made for the purpose of iterating over different systems (x86_64-linux, aarch64-linux, ...), and replacing it with a few lines of pure nix code (The `pkgsFor`).

You've already noticed one of the problems flake-utils has, it's not instantly clear what systems it iterates over. 
It also has several different other caveats, here's a nice blog post outlining some more of them: https://ayats.org/blog/no-flake-utils. 
You'll find that most bigger project with great nix support tend to remove flake-utils at some point, as it introduces more complexity to a flake.

PR #486, opened 2025-09-29
Add pause feature

MrNavaStar:
This PR adds the ability to pause the swww-daemon via the swww client. The command is `swww pause` and will toggle the paused state of the daemon. When paused, gpu usage drops to 0% which is nice for playing games.

This PR isn't very invasive and makes minimal changes, but let me know what you would like changed. Cheers!

### DIFF
--- a/daemon/src/main.rs
+++ b/daemon/src/main.rs
@@ -739,7 +739,7 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
         use rustix::event::{PollFd, PollFlags};
         use wayland::*;
         use WaylandObject::*;
-        
+
         daemon.backend.flush()?;
 
         let mut fds = [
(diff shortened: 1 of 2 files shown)


daemon/src/main.rs
```diff
@@ -752,6 +758,19 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
         let wayland_event = !fds[0].revents().is_empty();
         let socket_event = !fds[1].revents().is_empty();
 
+        if socket_event {
```
LGFae, line 761: This socket event cannot be moved up here or it will lead to ordering problems. Instead of putting `if daemon.paused { continue }`, you add this condition to the `daemon.draw()` at the end, so it becomes `if daemon.poll_time.is_some() && !daemon.paused { daemon.draw() }`

MrNavaStar:
I measured using `radeontop`. It's like htop for amd gpus. 

I have not tested that but I can. I'd still like this feature anyways as sometimes I just like to pause the gif cause it can look weird with my transparent terminal. 

I will add this to the docs today

daemon/src/main.rs
MrNavaStar: Ah I see, I'll fix that

LGFae:
There's a bizarre failure with `rustfmt`. It does not seem to matter, so I will be merging this as-is. Thanks!

